PR #28, opened 2021-10-28
Feat/login

ghost:
- Add auth directory for join, login and reset password flow
- Move join directory inside auth directory
- Change `JoinBox` to `AuthBox`
- Add `emailValidationRule` and `passwordValidationRule` as utils
- Add login page, forgot password apge and reset password page
- Add `AppAlert` component
- Add authentication handling


wpf500 (changes requested):
Nice work!

One bug though: when I try an invalid password it still redirects me to /profile, even though the call to /api/login returned a 400

wpf500 (approved):

PR #41, opened 2021-11-04
Theme page

wpf500:
A page where we can see the theme in all it's glory. We could build this out over time to include components etc.

image


ghost:
Looks good :+1:. A few notes that probably should be in documentation and/or checked by linting.

Component names should be PascalCase and multi words. (https://vuejs.org/v2/style-guide/#Multi-word-component-names-essential). For page components I'm using this convention: `ThemePage.vue`. Also I try to put space between tags  `<script>`, `<template>` and `<style>

In places I didn't follow these, I'm trying to refactor them.

Do you think it's necessary to have this as part of our bundle?

PR #48, opened 2021-11-18
Feat/contribution

ghost:


src/modules/auth/join/use-join.ts
```diff
@@ -32,7 +32,6 @@ const signUpData = reactive<SignUpData>({
 });
 
 const joinContent = ref<JoinContentData>({
-  currencySymbol: '',
```
ghost, line 35: @wpf500 We can remove this from API.

wpf500: :+1:

src/modules/contribution/use-contribution.ts
```diff
@@ -0,0 +1,234 @@
...
+  else if (isExpiringMember.value) return 'Restart contribution';
+  return 'Start contribution';
+});
+
+const showContributionForm = computed(() => {
+  return isAnnualMember.value && !isExpiredMember.value ? false : true;
+});
+
+const showCancelContribution = computed(() => {
+  return hasGoCardlessType.value || !isActiveMember.value ? false : true;
```
wpf500: This is very hard to read, the expression `hasGoCardlessType.value || !isActiveMember.value` is already a boolean so you shouldn't need to use a ternary if at all. I think the logic is wrong too, should be:
```js
return hasGoCardlessType.value && isActiveMember.value;
```

src/modules/contribution/use-contribution.ts
```diff
@@ -0,0 +1,234 @@
...
+
+const contributionButtonText = computed(() => {
+  if (hasManualPayment.value) return t('contribution.updatePaymentType');
+  else if (isActiveMember.value) return 'Update contribution';
+  else if (isExpiringMember.value) return 'Restart contribution';
+  return 'Start contribution';
+});
+
+const showContributionForm = computed(() => {
+  return isAnnualMember.value && !isExpiredMember.value ? false : true;
```
wpf500: Hard to read, this reads as: "is an annual member and not expired then negate", humans are not very good at negative logic (https://www.obsidiansecurity.com/blog/why-to-stop-writing-negative-code/)

Instead:
```js
return isExpiredMember.value || !isAnnualMember.value;
```
This reads as "is expired or not an annual member",

src/modules/contribution/components/PaymentSource.vue
```diff
@@ -0,0 +1,53 @@
+<template>
+  <div>
+    <div class="mb-4">
+      <FontAwesomeIcon
+        class="text-2xl -mb-1 mr-3"
+        :icon="['far', 'credit-card']"
+      />
+
+      <!-- - TODO: add dynamic content from API -->
```
wpf500: Remove comment

wpf500 (changes requested):
Good stuff! Just a couple of changes requested :)

src/modules/contribution/use-contribution.ts
```diff
@@ -0,0 +1,234 @@
...
+        paymentSource.accountHolderName = data.paymentSource.accountHolderName;
+        paymentSource.accountNumberEnding =
+          data.paymentSource.accountNumberEnding;
+      }
+    })
+    .catch((err) => err);
+};
+
+const newContribution = reactive<NewContribution>({
+  amount: 5,
```
wpf500: This should be set to the user's current contribution

src/modules/contribution/components/CancelContribution.vue
```diff
@@ -0,0 +1,42 @@
...
+    <p class="mb-3 text-sm">
+      {{
+        t('contribution.cancelDescription', {
+          day: formattedDate.day,
+          month: formattedDate.month,
+        })
+      }}
+    </p>
+
+    <AppButton variant="dangerOutlined" to="/contribution/cancel">
```
wpf500: URL should be `/profile/contribution/cancel`

wpf500 (approved):

PR #50, opened 2021-11-22
feat: add delivery address fields to setup page

ghost:


src/modules/auth/join/pages/SetupPage.vue
```diff
@@ -36,6 +36,17 @@
         />
       </div>
 
+      <template v-if="setupContent.showMailOptIn">
+        <AppAddress
```
wpf500, line 64: This also needs the mail opt in fields above it similar to the newsletter fields

So I guess it will be something like:
```vue
<p class="text-lg mb-1">
  {{ setupContent.mailTitle }}
</p>
<p class="mb-4 text-sm">
  {{ setupContent.mailText }}
</p>
<div class="mb-4">
  <input
    id="updates"
    v-model="memberData.profile.deliveryOptIn"
    type="checkbox"
    name="updates"
  />
  <label for="updates" class="font-bold ml-1">
    {{ setupContent.mailOptIn }}
  </label>
</div>
```

Here's an example from the current setup screen
image

wpf500: The address then becomes required if they tick the opt in

wpf500: Also note the checkbox needs to sync with `profile.deliveryOptIn` in /member/me

wpf500:
:+1:

PR #58, opened 2021-11-25
feat: add authorization for routes

ghost:


wpf500:
Code looks good but all of the routes we have so far don't need any role to be accessed. Users who have never had a membership or whose membership has expired will have an empty role list from /member/me:
```json
{
  ...,
  "roles": []
}
```

PR #59, opened 2021-11-25
Improvements for launch

wpf500:
- Add more loading states to buttons
- Add type submit to join forms
- Fix excess spacing on homepage

### DIFF
--- a/src/modules/notice/NoticeContainer.vue
+++ b/src/modules/notice/NoticeContainer.vue
@@ -1,5 +1,5 @@
 <template>
-  <section v-if="notices.length" class="mb-10 md:mb-12">
+  <section v-if="notices.length">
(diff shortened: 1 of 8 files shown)


src/modules/notice/NoticeContainer.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-  <div v-if="notices.length">
+  <section v-if="notices.length" class="mb-10 md:mb-12">
```
ghost: You can add this to `NoticeContainer` itself when it's used (in home page). In case this component is used somewhere else it might need different classes.
```
<NoticeContainer class="mb-10 md:mb-12">
```

wpf500: Ah cool, change applied!

ghost:
:+1:

PR #61, opened 2021-11-25
fix: routes meta pageTitle

ghost:


wpf500:
I think all of these need to be internationalised?

wpf500:
:+1:

PR #64, opened 2021-11-30
fix: make roles required in route meta

ghost:


wpf500:
@porivo I changed this because it doesn't seem to be enforced. When I did testing last week all of the endpoints like `/join/complete`, `/join/confirm-email/:id` etc. were broken because they have no `meta` at all. Unless `roles` existing can be enforced by the compiler I think the roles should be optional.

PR #84, opened 2022-01-20
Feat/footer

rlafuente:
This PR introduces a static footer for user and admin pages. 

Important details:
- Included is a fix to have the `canAdmin` helper function also return true when the user is a superadmin
- Right now the footer links are hardcoded in the template, we should later review how parts should be editable (or not)
- An email variable was added to the Vite env setup

Open questions:
- There's probably a more elegant way to include the footer in all pages
- ~~User/admin detection probably doesn't warrant a specific prop, the template itself should be aware of the user's role~~ this is fixed

TODO before merge:
- [x] Finish mobile layout
- [x] Integrate the "Thank you" message into the footer or at least have them play nice visually


src/utils/currentUserCan.ts
```diff
@@ -10,4 +10,5 @@ const currentUserCan = (role: Role): ComputedRef<boolean> => {
 
 export default currentUserCan;
 
-export const canAdmin = currentUserCan(Role.Admin);
+export const canAdmin =
```
wpf500: I thought I had fixed this on the server side (i.e. the API should always return admin as well if superadmin), did you find it not working?

src/modules/contribution/ContributionPage.vue
```diff
@@ -95,6 +95,8 @@
       ←{{ t('common.backToHome') }}
     </router-link>
   </div>
+
+  <Footer />
```
wpf500: As you guessed you can just put this into `layout/DashboardLayout.vue` so it shows on all dashboard pages

src/components/Footer.vue
```diff
@@ -0,0 +1,70 @@
...
+<script lang="ts" setup>
+import { canAdmin } from '../utils/currentUserCan';
+const newsroomName = import.meta.env.VITE_NEWSROOM_NAME;
+const newsroomLink = import.meta.env.VITE_NEWSROOM_LINK;
+const newsroomEmail = import.meta.env.VITE_NEWSROOM_EMAIL;
+</script>
+
+<style scoped>
+.footer {
+  @apply text-sm;
```
wpf500: Can `.footer` be removed in favour of just adding `.text-sm` directly above?

src/components/Footer.vue
```diff
@@ -0,0 +1,70 @@
...
+}
+
+a {
```
wpf500, line 86: This a bit risky as if someone removed `scoped` it would suddenly apply to the whole app, but obviously reduces a lot of duplication and is fine for now. Can you just add a comment above the selector just to remind any other developer who comes across it
```css
/* TODO: fix relying on scoped */
```

src/modules/home/HomePage.vue
```diff
@@ -74,7 +76,9 @@ import PageTitle from '../../components/PageTitle.vue';
 import AppButton from '../../components/forms/AppButton.vue';
 import AppAlert from '../../components/AppAlert.vue';
 import WelcomeMessage from '../../components/welcome-message/WelcomeMessage.vue';
+import Footer from '../../components/Footer.vue';
 import { useHome } from './use-home';
+import { canAdmin } from '../../utils/currentUserCan';
```
wpf500: Unused import

wpf500 (commented):
Looks really good in general, good stuff!

A few comments but just small things. The main thing is we need to add internationalisation, probably simpler for me to explain how we do this on a call.

Other outstanding thing as you say is to sort where the links go, I guess this needs to be done with the product team?

src/utils/currentUserCan.ts
rlafuente: Oh ok! I was looking into roles and noticed that function did not check for the SuperAdmin role, and thought it was an omission. We can throw this out if it's redundant

src/modules/contribution/ContributionPage.vue
rlafuente: Gotcha, thank you for pointing this out :)

src/utils/currentUserCan.ts
rlafuente: just confirmed that superadmin works as expected, I remember I changed this because I didn't see changes that I expected, but I suppose there was a different reason.

rlafuente:
Okay i think all code review points are addressed! I've been making single change commits for clarity since we can squash and merge at the end, but let me know if you'd prefer things done differently.

wpf500:
@rlafuente great, looks good. I definitely prefer individual commits :+1:. One comment remaining above about whether we need the `.footer` class or not then I think we should add `.hidden` to the footer and merge (so it goes live invisibly)

PR #94, opened 2022-02-09
Feat/customisable footer

wpf500:
Remove environment variables in favour of fetching content from API. Added a new `generalContent` to the store that preloads this content so it is available throughout the app


wpf500:
@rlafuente @aiscarvalho I've made a start here, basically making it so we can pull footer links (and other info) from the API.

`generalContent` in the store is preloaded with the response to the following API call:
https://dev.beabee.io/api/1.0/content/general

You should be able to use this to set the links and other info in the footer, as an example I've removed all the old environment variables we were using.

PR #97, opened 2022-02-10
Fix: profile left alignment

### LINKED ISSUE #88: Minor: misaligned title in user home, in viewport interval 656px up to 767px
In the user homepage, between 656px and 767px wide, the profile summary title is left aligned.
It should be the same as the title above it (quick actions).

image

basilesimon:
Hopefully address #88 

https://user-images.githubusercontent.com/5701152/153442568-60b6f1b3-ce67-4a8f-84b1-2256d4126144.mov

### DIFF
--- a/src/modules/home/HomePage.vue
+++ b/src/modules/home/HomePage.vue
@@ -4,7 +4,7 @@
       <font-awesome-icon class="transform -rotate-90" :icon="['fas', 'key']" />
     </template>
 
-    {{ t('homePage.passwordReset') }}
+    {{ `${t('homePage.passwordReset')}` }}
   </AppAlert>
   <!-- TODO: check page title implementation, after implementing other pages
     you might want to improve it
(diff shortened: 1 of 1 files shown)


src/modules/home/HomePage.vue
```diff
@@ -4,7 +4,7 @@
       <font-awesome-icon class="transform -rotate-90" :icon="['fas', 'key']" />
     </template>
 
-    {{ t('homePage.passwordReset') }}
+    {{ `${t('homePage.passwordReset')}` }}
```
wpf500: I don't think this is needed?

wpf500 (changes requested):
Looks good apart from what looks like an accidental commit :+1:, can you revert that line?

src/modules/home/HomePage.vue
basilesimon: Mmmh, funny things are happening with the linter/fixer triggering upon committing. I managed to revert it – will let you mark this as resolved.

wpf500 (approved):

src/modules/home/HomePage.vue
wpf500: Yeah we should definitely review whether the linter/fixer triggering on commit is useful or not, I find it a bit jarring

PR #100, opened 2022-02-11
Proposal: fix startup warning on dynamic import

basilesimon:
This PR removes this warning on startup... though does not address it.

```sh
3:46:21 PM [vite] warning: 
/Users/b/github/beabee-frontend/src/layouts/AppLayout.vue
11 |        try {
12 |          const layoutName = meta.layout ? meta.layout : "Dashboard";
13 |          const component = await import(`./${layoutName}Layout.vue`);
   |                                         ^
14 |          layout.value = component.default;
15 |        } catch (e) {
The above dynamic import cannot be analyzed by vite.
See https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#limitations for supported dynamic import formats. If this is intended to be left as-is, you can use the /* @vite-ignore */ comment inside the import() call to suppress this warning.
```

To conform to what `vite` requires, we'd do:

```js
const component = await import(`./${layoutName}Layout.vue`);
// to
const component = await import(`./layout-${layoutName}.vue`);
```

and rename the files in `src/layouts/` appropriately.

Happy to discuss!


wpf500:
@basilesimon if you want to try fixing this error properly by renaming the layout files feel free, would be good to fix this warning :)

basilesimon:
@wpf500 I've done so – could you run a cursory test of these layouts, please?

wpf500 (approved):

PR #102, opened 2022-02-11
Callouts page

wpf500:
Notes:
* I changed the filter from unanswered to answered, I'm not sure what the use case would be for filtering for archived callouts that you can no longer answer, but you might want to look back at the ones you did answer
* I added pagination as the list of archived callouts could be quite large, at the moment it shows 15 callouts per page
* Some archived callouts don't have an end date because historically they didn't have to have one. However with the changes agreed in https://github.com/beabee-communityrm/beabee/issues/111 all callouts will have an end date in the future, so this quirk can be ignored

Screenshots:

image

image

### DIFF
--- a/src/components/forms/AppSearchInput.vue
+++ b/src/components/forms/AppSearchInput.vue
@@ -1,0 +1,31 @@
+<template>
+  <form
+    class="relative"
+    @submit.prevent="emit('update:modelValue', searchText)"
+  >
+    <AppInput v-model="searchText" class="pr-8" :placeholder="placeholder" />
+    <button class="absolute right-0 w-8 h-full">
+      <font-awesome-icon icon="search" />
+    </button>
+  </form>
+</template>
+
+<script lang="ts" setup>
+import { ref, watch } from 'vue';
+import AppInput from './AppInput.vue';
+
+const emit = defineEmits(['update:modelValue']);
+const props = defineProps<{
+  modelValue: string;
+  placeholder: string;
+}>();
+
+const searchText = ref(props.modelValue);
+
+watch(
+  () => props.modelValue,
+  (newValue) => {
+    searchText.value = newValue;
+  }
+);
+</script>
--- a/src/modules/contacts/components/SearchBox.vue
+++ b/src/modules/contacts/components/SearchBox.vue
@@ -1,19 +1,7 @@
 <template>
   <div class="flex">
     <!--<div class="flex-1">actions</div>-->
-    <form
-      class="relative"
-      @submit.prevent="emit('update:modelValue', basicSearchValue)"
-    >
-      <AppInput
-        v-model="basicSearchValue"
-        class="pr-8"
-        :placeholder="t('contacts.search')"
-      />
-      <button class="absolute right-0 w-8 h-full">
-        <font-awesome-icon icon="search" />
-      </button>
-    </form>
+    <AppSearchInput v-model="searchText" :placeholder="t('contacts.search')" />
     <a
       href="/members?type=advanced"
       class="ml-2 p-2"
@@ -28,9 +16,9 @@
 </template>
 
 <script lang="ts" setup>
-import { ref, watch } from 'vue';
+import { computed, ref } from 'vue';
 import { useI18n } from 'vue-i18n';
-import AppInput from '../../../components/forms/AppInput.vue';
+import AppSearchInput from '../../../components/forms/AppSearchInput.vue';
 
 const { t } = useI18n();
 
@@ -39,12 +27,9 @@ const props = defineProps<{
   modelValue: string;
 }>();
 
-const basicSearchValue = ref(props.modelValue);
-watch(
-  () => props.modelValue,
-  (value) => {
-    basicSearchValue.value = value;
-  }
-);
+const searchText = computed({
+  get: () => props.modelValue,
+  set: (text) => emit('update:modelValue', text),
+});
 const showAdvancedSearch = ref(false);
 </script>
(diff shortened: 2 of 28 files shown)


src/utils/api/callout.ts
```diff
@@ -9,13 +14,25 @@ function toDate(s: string | undefined): Date | undefined {
...
   });
-  return data.map((callout) => ({
-    ...callout,
-    starts: toDate(callout.starts),
-    expires: toDate(callout.expires),
-  }));
+
+  return {
+    ...data,
+    items: data.items.map(toCallout),
```
basilesimon, line 36: The type checker is complaining about `data.items` being `undefined`.
We're not typing `data` here but I believe it comes in as this kind?

```ts
{
  slug: string;
  title: string
  excerpt: string;
  image?: string;
  starts?: string;
  expires?: string;
}
```

src/utils/api/api.interface.ts
```diff
@@ -165,13 +171,24 @@ export interface ProfileContent {
   introMessage: string;
 }
 
-export interface BasicCalloutData {
+export interface GetBasicCalloutData {
```
basilesimon, line 186: I'd like us to discuss this prefixing again, actually. Reading the implementations of this interface, I now feel the `Get-` prefix is an indirection:

`fn = query: GetCalloutsQuery => BasicCalloutData` makes more sense to me.

Also, if this interface is "basic", what is its non-basic version?

src/components/forms/AppSearchInput.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <form
+    class="relative"
+    @submit.prevent="emit('update:modelValue', searchText)"
+  >
+    <AppInput v-model="searchText" class="pr-8" :placeholder="placeholder" />
```
basilesimon, line 6: Add `:inputType="text"` as it's not the default on this component?

src/modules/contacts/components/SearchBox.vue
```diff
@@ -39,12 +27,9 @@ const props = defineProps<{
...
-watch(
-  () => props.modelValue,
-  (value) => {
-    basicSearchValue.value = value;
-  }
-);
+const searchText = computed({
+  get: () => props.modelValue,
+  set: (text) => emit('update:modelValue', text),
+});
```
basilesimon, line 33: Would love to talk about `watch()` v `computed()` – I remember reading about it, I'd be curious to hear why you changed this.

src/modules/home/HomePage.vue
```diff
@@ -113,14 +114,27 @@ const profileContent = ref<ProfileContent>({
...
-  callouts.value = await fetchCallouts();
+  callouts.value = (
+    await fetchCallouts({
+      rules: {
+        condition: 'AND',
+        rules: [
+          {
+            field: 'status',
+            operator: 'equal',
+            value: CalloutStatus.Open,
```
basilesimon, line 133: Sweet 👍

src/modules/contacts/components/SearchBox.vue
wpf500: In this case it's just because I moved the `watch()` into `AppSearchInput`, so this component now just proxies the `modelValue` using a `computed()`

src/modules/home/HomePage.vue
wpf500: It's a bit verbose at the moment but the general idea is to use a standard query language for all resources on the API (where it makes sense). I'm interested if you have any thoughts on this, I looked around to see what sort of standards there were but couldn't see anything, although I found [FIQL](https://fiql-parser.readthedocs.io/en/stable/usage.html) interesting

src/components/forms/AppSearchInput.vue
wpf500: Ah I didn't realise, thanks. It feels like the correct behaviour would be that it defaults to `type="text"` as that would reflect the real HTML `input` element. What do you think?

src/utils/api/api.interface.ts
wpf500: Yeah definitely up for chatting about this as I agree it does seem a bit pointless. I added a `Get` prefix so that both `Get` and `Update` could derive from the same base class (non-prefixed). 

```typescript
interface Item {
  id: string;
  // other shared properties
}

interface GetItem extends Item {
  created: Date;
  // other read only properties
}

interface UpdateItem extends Item {
  // write only properties??
}
```

Maybe a better route would be `Item` => `BaseItem`, `GetItem` => `Item`?

On the `Basic` vs `More`, this is to reflect that you get more data back from a request to a specific resource than when you load a list of resources.
e.g.
`GET /api/1.0/callout` loads a list of callouts with basic info
`GET /api/1.0/callout/specific-callout` loads a specific callout with more info
Again open to discussing what the right convention is

src/components/forms/AppSearchInput.vue
basilesimon: I feel the same way!

src/utils/api/api.interface.ts
basilesimon: Yeah, I think I understand your motivation behind separating, extending and prefixing the types.

For now, my point is solely about terminology. Function names which include a verb lead you to guess what they do, and so I wonder if this signpost belongs on the types as well.

In this instancem this is the signature of a function using this type:

```ts
fn fetchCallouts = query?: GetCalloutsQuery => Promise<Paginated<GetBasicCalloutData>>
```

`Get-` prefix feels right on the function parameter ; but dissonant on the return.

PR #104, opened 2022-02-22
Contact overview page for admins

basilesimon:
e.g. [http://localhost:3000/contacts/0adfc2c8-c35f-4ad3-8b4a-3118a74f7797](http://localhost:3000/contacts/0adfc2c8-c35f-4ad3-8b4a-3118a74f7797)

based on [🎨 single contact designs](https://design.penpot.app/#/workspace/887044a0-9c39-11eb-b4cd-81aaa3a367f9/3342c4a0-46c8-11ec-8508-6194167bb409?page-id=224b7680-576f-11ec-bc54-4148703305a8)

--

Todo list:
- [x] Clarify what "roles" are: an array of dates?
- [x] Implement "Security" section
- [x] Include default "-" when no option: out of this PR's scope
- [x] Localise [this para](https://github.com/beabee-communityrm/beabee-frontend/pull/104/files#diff-9ca9ac60d0dd5675df8fef4e5a2c075630d5c7659c63a9f501e30f6adcdbf54aR68-R72)
- [x] The "About" form should support updating the contact/member: `updateMember`
- [x] Switch list redirect from `/members/:id` to `/contacts/:id`
- [x] Re-implement grid layout to make sections close to each other vertically
- [x] Add an `AppTextArea` component?

### DIFF
--- a/src/components/forms/AppTextArea.vue
+++ b/src/components/forms/AppTextArea.vue
@@ -1,9 +1,8 @@
 <template>
-  <label v-if="label" class="block mb-1.5 font-semibold" :for="inputType"
+  <label v-if="label" class="block mb-1.5 font-semibold"
     >{{ formattedLabel }}
   </label>
   <textarea
-    :id="inputType"
     class="
       p-2
       w-full
@@ -11,7 +10,6 @@
       rounded
       focus:outline-none focus:shadow-input
     "
-    :type="inputType"
     :class="dangerClasses"
     :value="modelValue"
     v-bind="$attrs"
@@ -40,7 +38,6 @@ import InfoMessage from '../InfoMessage.vue';
 const props = withDefaults(
   defineProps<{
     modelValue?: string;
-    inputType?: 'text';
     label?: string;
     errorMessage?: string | Ref<string>;
     infoMessage?: string;
@@ -48,7 +45,6 @@ const props = withDefaults(
   }>(),
   {
     modelValue: '',
-    inputType: 'text',
     label: undefined,
     errorMessage: undefined,
     infoMessage: undefined,
--- a/src/components/forms/AppTextArea.vue
+++ b/src/components/forms/AppTextArea.vue
@@ -1,0 +1,68 @@
+<template>
+  <label v-if="label" class="block mb-1.5 font-semibold" :for="inputType"
+    >{{ formattedLabel }}
+  </label>
+  <textarea
+    :id="inputType"
+    class="
+      p-2
+      w-full
+      border border-primary-40
+      rounded
+      focus:outline-none focus:shadow-input
+    "
+    :type="inputType"
+    :class="dangerClasses"
+    :value="modelValue"
+    v-bind="$attrs"
+    :required="required"
+    @input="$emit('update:modelValue', handleInput($event))"
+  />
+
+  <div
+    v-if="errorMessage"
+    class="text-xs text-danger font-semibold mt-1.5"
+    role="alert"
+  >
+    {{ errorMessage }}
+  </div>
+
+  <div v-if="infoMessage" class="mt-2 text-xs">
+    <InfoMessage :message="infoMessage" />
+  </div>
+</template>
+
+<script lang="ts" setup>
+import { computed, Ref } from '@vue/reactivity';
+import handleInput from '../../utils/handle-input';
+import InfoMessage from '../InfoMessage.vue';
+
+const props = withDefaults(
+  defineProps<{
+    modelValue?: string;
+    inputType?: 'text';
+    label?: string;
+    errorMessage?: string | Ref<string>;
+    infoMessage?: string;
+    required?: boolean;
+  }>(),
+  {
+    modelValue: '',
+    inputType: 'text',
+    label: undefined,
+    errorMessage: undefined,
+    infoMessage: undefined,
+    required: false,
+  }
+);
+
+defineEmits(['update:modelValue']);
+
+const dangerClasses = computed(() => {
+  return props.errorMessage ? ['bg-danger-10', 'border-danger-70'] : null;
+});
+
+const formattedLabel = computed(() => {
+  return props.required ? props.label + '*' : props.label;
+});
+</script>
--- a/src/modules/contacts/pages/ContactOverviewPage.vue
+++ b/src/modules/contacts/pages/ContactOverviewPage.vue
@@ -110,11 +110,11 @@
       <AppInfoList v-for="role in contact.roles" :key="role.role">
         <AppInfoListItem :name="t(`common.role.${role.role}`)">
           <AppRoundBadge :type="role.dateExpires ? 'warning' : 'success'" />
-          {{ formatLocale(role.dateAdded, 'P') }}
+          {{ formatLocale(role.dateAdded, 'P') + ' → ' }}
           {{
             role.dateExpires
-              ? ' → ' + formatLocale(role.dateExpires, 'P')
-              : ' onwards'
+              ? formatLocale(role.dateExpires, 'P')
+              : t('contacts.data.rolesCopy.today')
           }}
         </AppInfoListItem>
       </AppInfoList>
--- a/src/modules/contacts/pages/ContactOverviewPage.vue
+++ b/src/modules/contacts/pages/ContactOverviewPage.vue
@@ -71,7 +71,7 @@
         />
       </AppInfoList>
     </div>
-    <div class="row-span-3">
+    <div class="row-span-3 max-w-xl">
       <AppHeading>{{ t('contactOverview.about') }}</AppHeading>
       <p class="mb-5">
         {{ t('contactOverview.annotation.copy.begin') }}
@@ -105,8 +105,7 @@
       <AppHeading>{{ t('contactOverview.roles') }}</AppHeading>
       <AppInfoList v-for="role in contact.roles" :key="role.role">
         <AppInfoListItem :name="t(`common.role.${role.role}`)">
-          <!-- @TODO: make component -->
-          <span class="inline-block w-2 h-2 bg-success rounded-full"></span>
+          <AppRoundBadge :type="role.dateExpires ? 'warning' : 'success'" />
           {{ formatLocale(role.dateAdded, 'P') }}
           {{
             role.dateExpires
@@ -151,6 +150,7 @@ import AppHeading from '../../../components/AppHeading.vue';
 import AppInput from '../../../components/forms/AppInput.vue';
 import AppTextArea from '../../../components/forms/AppTextArea.vue';
 import AppButton from '../../../components/forms/AppButton.vue';
+import AppRoundBadge from '../../../components/AppRoundBadge.vue';
 import { onBeforeMount, ref, reactive } from 'vue';
 import { ContributionType } from '../../../utils/enums/contribution-type.enum';
 import {
--- a/src/modules/contacts/pages/ContactOverviewPage.vue
+++ b/src/modules/contacts/pages/ContactOverviewPage.vue
@@ -70,22 +70,26 @@
         <b>{{ t('contactOverview.annotation.copy.bold') }}</b>
         {{ t('contactOverview.annotation.copy.end') }}
       </p>
-      <!-- @submit.prevents="" -->
-      <form>
+
+      <form @submit.prevent="handleFormSubmit">
         <AppInput
-          v-model="notesStub"
+          v-model="contactAnnotations.notes"
           input-type="text"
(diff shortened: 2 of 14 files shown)


basilesimon:
"Include default "-" when no option"

@wpf500 what do you recommend in this instance?
I could probably `compute()` UI-friendly strings, instead of relying on `contact`'s raw data fields.

basilesimon:
re: https://github.com/beabee-communityrm/beabee-frontend/pull/104/commits/0ff598f9579a42652b120e60e5e9f375160976ff

how do I stop the whole app redrawing? you mentioned a magic component... 😉

wpf500:
> "Include default "-" when no option"
> 
> @wpf500 what do you recommend in this instance? I could probably `compute()` UI-friendly strings, instead of relying on `contact`'s raw data fields.

I think for this I would opt to handle it in the `AppInfoListItem` component, so if it gets an empty value it defaults to "-"

basilesimon:
@aiscarvalho - @aiscarvalho 

There are several design changes that came up during the initial implementation of this page. Here's the overview:

image

### On Roles

image

- I've added a unicode arrow in lieu of the dash separating the dates. Hope that's fine!
- The little green badge is using our "success" colour if the role is current, "warning" otherwise. As intended?
- For current roles (see "superadmin" above), what do we want to display? I added "onwards" but that doesn't feel like a good candidate for localisation.

### On Security

https://user-images.githubusercontent.com/5701152/155365510-486dfc0c-caa0-43b1-a5a9-a1b63f6989d6.mov

- Will and I changed the layout slightly, it felt a bit more natural. Could you review?
- Clicking one of the buttons disables both and displays the result in a read-only input field below. How's that?

src/modules/contacts/pages/ContactOverviewPage.vue
```diff
@@ -17,34 +17,196 @@
...
+          :name="t('contacts.data.email')"
+          :value="contact.email"
+        />
+        <AppInfoListItem
+          :name="t('contacts.data.phone')"
+          :value="contact.profile.telephone"
+        />
+        <AppInfoListItem
+          :name="t('contacts.data.deliveryOptIn')"
+          :value="contact.profile.deliveryOptIn"
```
wpf500: This should be internationalised using `common.yes` and `common.no`

src/modules/contacts/pages/ContactOverviewPage.vue
```diff
@@ -17,34 +17,196 @@
...
+          "
+        />
+        <AppInfoListItem
+          :name="t('contacts.data.period')"
+          :value="contact.contributionPeriod"
+        />
+        <AppInfoListItem
+          v-if="contact.contribution.type === ContributionType.GoCardless"
+          :name="t('contacts.data.payingFee')"
+          :value="contact.contribution.payFee"
```
wpf500: This should be internationalised using common.yes and common.no

src/modules/contacts/pages/ContactOverviewPage.vue
```diff
@@ -17,34 +17,196 @@
...
+        >
+      </form>
+    </div>
+
+    <div>
+      <AppHeading>{{ t('contactOverview.roles') }}</AppHeading>
+      <AppInfoList v-for="role in contact.roles" :key="role.role">
+        <AppInfoListItem :name="t(`common.role.${role.role}`)">
+          <!-- @TODO: make component -->
+          <span class="inline-block w-2 h-2 bg-success rounded-full"></span>
```
wpf500: Make this into a little component which takes a colour as an attribute

src/modules/contacts/pages/ContactOverviewPage.vue
```diff
@@ -17,34 +17,196 @@
...
+      </AppInfoList>
+    </div>
+    <div>
+      <AppHeading>{{ t('contactOverview.security.title') }}</AppHeading>
+      <p>{{ t('contactOverview.security.whatDoTheButtonsDo') }}</p>
+      <form @submit.prevent="handleSecurityAction">
+        <AppButton
+          type="submit"
+          variant="link"
+          :disabled="!securityButtonsEnabled"
```
wpf500: I'm not a huge fan of inverse logic where possible, could this be more readable as `securityButtonsDisabled` instead?

src/components/forms/AppTextArea.vue
```diff
@@ -0,0 +1,68 @@
...
+  <textarea
+    :id="inputType"
+    class="
+      p-2
+      w-full
+      border border-primary-40
+      rounded
+      focus:outline-none focus:shadow-input
+    "
+    :type="inputType"
```
wpf500: I don't think we need a type attribute for this component as it's not relevant to a `<textarea>`

src/components/forms/AppTextArea.vue
```diff
@@ -0,0 +1,68 @@
+<template>
+  <label v-if="label" class="block mb-1.5 font-semibold" :for="inputType"
```
wpf500: Hmm this isn't related to your task but I'm not sure how the `<label for="inputType">` and `<input id="inputType">` ever worked here, this clearly won't generate unique IDs! I'm guessing this means all our labels are broken if you try to use them to focus an input. I'll add an issue to fix this.

src/modules/contacts/pages/ContactOverviewPage.vue
```diff
@@ -17,34 +17,196 @@
...
+          :name="t('contacts.data.deliveryOptIn')"
+          :value="contact.profile.deliveryOptIn"
+        />
+        <AppInfoListItem
+          :name="t('contacts.data.newsletter')"
+          :value="contact.profile.newsletterStatus"
+        />
+        <AppInfoListItem
+          :name="t('contacts.data.groups')"
+          :value="contact.profile.newsletterGroups.join(',')"
```
wpf500: Join with `', '` (a space after the comma) for slightly nicer formatting?

wpf500 (commented):
Looks great, nice work :+1:. A few comments here and there and some stuff left to do, but mostly because you are waiting on me to create the API calls ;)

src/modules/contacts/pages/ContactOverviewPage.vue
basilesimon: Done: https://github.com/beabee-communityrm/beabee-frontend/pull/104/commits/5e4f54269287cd70ffe74ac20a25b3c328a4b585

src/modules/contacts/pages/ContactOverviewPage.vue
basilesimon: Done: https://github.com/beabee-communityrm/beabee-frontend/pull/104/commits/578f343a4e4679ed35823e36a989aa613b9789b8

src/components/forms/AppTextArea.vue
basilesimon: Done: https://github.com/beabee-communityrm/beabee-frontend/pull/104/commits/578f343a4e4679ed35823e36a989aa613b9789b8

src/components/forms/AppTextArea.vue
basilesimon: Done: https://github.com/beabee-communityrm/beabee-frontend/pull/104/commits/578f343a4e4679ed35823e36a989aa613b9789b8

src/modules/contacts/pages/ContactOverviewPage.vue
basilesimon: Done: https://github.com/beabee-communityrm/beabee-frontend/pull/104/commits/578f343a4e4679ed35823e36a989aa613b9789b8

src/modules/contacts/pages/ContactOverviewPage.vue
basilesimon: Done: https://github.com/beabee-communityrm/beabee-frontend/pull/104/commits/578f343a4e4679ed35823e36a989aa613b9789b8

src/modules/contacts/pages/ContactOverviewPage.vue
basilesimon: Done: https://github.com/beabee-communityrm/beabee-frontend/pull/104/commits/578f343a4e4679ed35823e36a989aa613b9789b8

basilesimon:
> Looks great, nice work 👍. A few comments here and there and some stuff left to do, but mostly because you are waiting on me to create the API calls ;)

Thanks for the review, I've addressed all of your points.

aiscarvalho:
>     * I've added a unicode arrow in lieu of the dash separating the dates. Hope that's fine!

yes!

>     * The little green badge is using our "success" colour if the role is current, "warning" otherwise. As intended?

Apologies because we didn't document this in the design.
The coloured dot is related to the member role only, it's an indication of an active (=paying) or expired (=non-paying) member.
So now that you mentioned it we realise it can be read as related to the role duration.  Maybe it makes more sense to move it to the "Contribution" section instead of here. This way we don't need to have a visual element that only exists for this role.

Should we mock that change in the design?
 
>     * For current roles (see "superadmin" above), what do we want to display? I added "onwards" but that doesn't feel like a good candidate for localisation.

We'll get back to this tomorrow!
 
>     * Will and I changed the layout slightly, it felt a bit more natural. Could you review?

Looks great!

wpf500:
> Apologies because we didn't document this in the design.
> The coloured dot is related to the member role only, it's an indication of an active (=paying) or expired (=non-paying) member.
> So now that you mentioned it we realise it can be read as related to the role duration. Maybe it makes more sense to move it to the "Contribution" section instead of here. This way we don't need to have a visual element that only exists for this role.

I think the coloured dot should actually apply to the role for now. There is a lot of confusion between the member role and having an active contribution which we definitely need to revisit, but I don't think it should block this task. So for now all let's just keep this the same as in the legacy app: all roles have a status, status is:
```
green = dateAdded < now && (!dateExpires || dateExpires > now), red otherwise.
```

rlafuente:
Picking up the open point about date ranges with no end date, using "onwards": 

Not sure whether there's a standard way to signal this, and personally I'd go for "2022/02/24 –⁠ present", but it's more of an editorial/copy decision that should take translation into account (good point @basilesimon). "Present" would probably be more translation-friendly, but it is understandable as it is and shouldn't constitute a blocker for this issue.

All good to go from our side :ok_hand:

basilesimon:
> Not sure whether there's a standard way to signal this, and personally I'd go for "2022/02/24 –⁠ present", but it's more of an editorial/copy decision that should take translation into account

I love it – implemented it in commit above, thanks for your help.

basilesimon:
@wpf500 I implemented your logic to decide which colour to render the badge in https://github.com/beabee-communityrm/beabee-frontend/pull/104/commits/5cc57adb2bcb9607193eebb2efbde05a0d2581e9 – can you review?

basilesimon:
I re-targeted this PR, which will now land into `dev`

aiscarvalho:
Looks good!

We've had a first view of the page and found the following items:
- in mobile, the overview list is mis-aligned with the others
- the save is not yet working: we wrote in the notes and description, saved and shifted to tab information; when we returned to overview notes and description were empty again
- the "save button" should be primary color instead of link
- the security section is not visible yet

image

basilesimon:
Here's the fix I can offer you, given that email addresses are arbitrarily long:

Screenshot 2022-03-01 at 10 27 36

Screenshot 2022-03-01 at 10 27 52

Will address **link colour** in a coming commit.

I left the **security section** hidden because we're missing the API side, so I'll turn this on in the future (you can toggle the CSS class in the meantime).

And about **saving the notes**, this is the meaning of my earlier message:

> The about section, which lets you annotate a contact, needs a bit of UX thinking, I believe. Empty textboxed make sense when there's no data yet, but what should we present when someone has notes written about them already?

aiscarvalho:
The alternative summary layout, for mobile, looks great! Very tidy and aligned now.
For ultra long email addresses we can break the line: the label first and the email below (if that makes sense to tackle).

Editorial: could we suggest using an em dash for empty fields?

The security section looks fine but the button style is off. These should be outlined buttons.

I had replied about **saving the notes** in Trello, so copying it here:
 
> thanks for bringing that up, it was something we [discussed over the design spec](https://trello.com/c/2htyLaCL/333-design-elements-contact-overview-page-%F0%9F%8E%A8).
>
> Our first take for the contact overview page was to show a comprehensive summary, and nothing was editable in this page but in its specific page. Will and Julia made some fair arguments over the need for quick note taking and so we kept the fields for notes, description and tags editable in this page. If there is information in these textboxes it should be editable right away, as it is in the legacy app (and the static spec :)

Almost ready to close!

PR #111, opened 2022-02-24
Contact payment API

wpf500:


src/utils/api/member.ts
```diff
@@ -158,3 +160,17 @@ export async function completeUpdatePaymentSource(
   );
   return toContrib(data);
 }
+
+export async function fetchPayments(
+  query: GetPaymentsQuery
+): Promise<Paginated<GetPaymentData>> {
+  const { data } = await axios.get('/member/me/payment', { params: query });
```
basilesimon:
```js
/member/${id}/payment
```
?

basilesimon (changes requested):

src/utils/api/member.ts
basilesimon: with obvs `id` being added to params.

src/utils/api/member.ts
```diff
@@ -158,3 +160,17 @@ export async function completeUpdatePaymentSource(
   );
   return toContrib(data);
 }
+
+export async function fetchPayments(
```
basilesimon, line 164: I've been wondering about renaming this to `fetchContributionHistory`.... but that wouldn't be accurate, would it

wpf500: I can see the logic to this but I think it would need to be part of a wider renaming job so will leave as `fetchPayments` for now

src/utils/api/member.ts
wpf500: :+1:

basilesimon (approved):
👍

PR #113, opened 2022-02-24
WIP: payments history table

basilesimon:
This PR implements a new component allowing a member to see the history of their financial contributions when visiting `/profile/contribution`.

Screenshots and original designs are below.

Please note that a couple of things (pagination buttons, table rows) look ever so slightly different from the designs – that is, pardon the pun, _by design._ I chose to favour consistency and our current implementation over including new `if/else` statements on otherwise quite standardised components.

Ana, Ricardo, I'll trust you'll flag what matters to you too 🙂

**One item of note** is the omission of the "select year" button, which requires a new round trip to the API while we've gone to some lengths to cut the size of the queries with pagination, as well as some other on-the-fly minor computations. For the sake of moving the larger piece of work across the line, I've left this out and would prefer it landed in a backlog of small tasks I can pick up when there's a bit more time, if that's okay with you.

---

### Original design
image

### DIFF
--- a/src/modules/contribution/components/PaymentsHistory.vue
+++ b/src/modules/contribution/components/PaymentsHistory.vue
@@ -14,11 +14,13 @@
         ><b>{{ n(value, 'currency') }}</b></template
       >
     </AppTable>
-    <AppPagination
-      v-model="currentPage"
-      :total-pages="totalPages"
-      class="mt-6"
-    />
+    <div class="flex w-full justify-center">
+      <AppPagination
+        v-model="currentPage"
+        :total-pages="totalPages"
+        class="mt-6"
+      />
+    </div>
   </div>
 </template>
 
--- a/src/modules/contribution/components/PaymentsHistory.vue
+++ b/src/modules/contribution/components/PaymentsHistory.vue
@@ -36,7 +36,7 @@ import { Paginated } from '../../../utils/api/api.interface';
 import { GetPaymentData } from '../../../utils/api/api.interface';
 import { Header, SortType } from '../../../components/table/table.interface';
 
-const { n } = useI18n();
+const { t, n } = useI18n();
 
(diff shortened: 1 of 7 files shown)


basilesimon:
Screen Shot 2022-03-01 at 14 56 13
Screen Shot 2022-03-01 at 14 56 18
Screen Shot 2022-03-01 at 14 56 46

src/modules/contribution/components/PaymentsHistory.vue
```diff
@@ -0,0 +1,78 @@
+<template>
+  <div>
+    <SectionTitle class="mb-2">Payment history </SectionTitle>
```
wpf500: Needs to be localised

wpf500 (changes requested):
Nice one :+1:, just that one string to localise

src/modules/contribution/components/PaymentsHistory.vue
basilesimon: nice one thanks

aiscarvalho:
Nice work! Things we noticed:
- the amount value, in the payment table, is too boldy. It is set in font-weight 900 and should be 700
- could we set a minimum table height of 12 rows (=12 payments) to avoid the jumpy pagination?
- pagination is missing the hover -- this is a general one, so maybe we add an issue to cover it?

We had a look but didn't quite understand what you meant by "favouring consistency" in the table. We can chat through it next week to better understand your vision :)

It's perfectly fine that we pick the select later. We can take some time to flesh out how the select and the pagination interact with each other, that isn't so clear atm.

basilesimon:
Addressed all concerns above!

https://user-images.githubusercontent.com/5701152/156747173-9b656150-d226-450c-a9a9-c2e7dbdabe2b.mov

PR #115, opened 2022-03-01
Review of the contact overview

basilesimon:
- Fix mobile layout of some items
- Display notes... when they've previously been set

### DIFF
--- a/src/modules/contacts/pages/ContactOverviewPage.vue
+++ b/src/modules/contacts/pages/ContactOverviewPage.vue
@@ -10,7 +10,7 @@
         <AppInfoListItem
           :name="t('contacts.data.lastSeen')"
           :value="
-            contact.lastSeen ? formatLocale(contact.lastSeen, 'PPP') : '-'
+            contact.lastSeen ? formatLocale(contact.lastSeen, 'PPP') : '–'
           "
         />
       </AppInfoList>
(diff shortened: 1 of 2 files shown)


src/modules/contacts/pages/ContactOverviewPage.vue
```diff
@@ -224,6 +226,8 @@ onBeforeMount(async () => {
     'contribution',
     'roles',
   ]);
+  contactAnnotations.notes = contact.value.profile.notes as string;
```
wpf500: I think this should be defaulted to empty string rather than casting `string | undefined` to `string`;
e.g.
```ts
contactAnnotation.notes = contact.value.profile.notes || '';
```

src/modules/contacts/pages/ContactOverviewPage.vue
```diff
@@ -224,6 +226,8 @@ onBeforeMount(async () => {
...
+  contactAnnotations.description = contact.value.profile.description as string;
```
wpf500: Same as above

wpf500 (changes requested):

basilesimon:
@wpf500 yes, pretty heavy-handed of me, sorry – now fixed!

wpf500 (approved):

basilesimon:
@aiscarvalho, I've addressed [your feedback](https://github.com/beabee-communityrm/beabee-frontend/pull/104#issuecomment-1056789253) in the above commits and I'm moving this to the next step now :)

PR #119, opened 2022-03-02
Feat/callout page

wpf500:
The new callout response page!

Notes:
* I haven't implemented the sidebar changes, currently it shows the normal sidebar for logged in users and nothing for guests. I'd like to defer these changes to a later date as I'm not clear on the functionality
  * Callouts link in sidebar is still hidden, but will be made visible before this goes live to clients
* Share button does nothing, probably just hide for now?

For testing you can use one callout that requires membership and one that doesn't:

This callout is only for members
https://dev.beabee.io/callouts/agm2020-feedback

This one allows guests
https://dev.beabee.io/callouts/agm2020-feedback-guest

## Screenshots

Lots and lots of different variations of this page:

Logged in user, not answered
image

Logged in user, thank you for responding
image

Logged in user, has previously responded
image

Non-logged in user, callout allows guest responses
image

Non-logged in user, callout only for members
image


aiscarvalho:
It looks beautiful!

We can only add one minor suggestion, that is not a blocker:   
we don't need to show the (disabled) submit button in _response view mode_
image

wpf500:
Reopening this PR as I think it was accidentally closed.

Agreed @aiscarvalho on the button but unfortunately this is currently controlled by the Form.IO plugin so not easy to change at the moment

src/utils/api/api.interface.ts
```diff
@@ -214,6 +218,38 @@ export interface GetCalloutsQuery
...
+
+export type GetCalloutResponsesQuery = GetPaginatedQuery<'member'>;
+
+type CalloutResponseAnswer =
+  | string
+  | boolean
+  | number
+  | null
+  | undefined
+  | Record<string, boolean>;
```
basilesimon, line 238: lol

src/modules/callouts/components/GuestFields.vue
```diff
@@ -0,0 +1,75 @@
...
+}>();
+
+const nameProxy = computed({
+  get: () => props.name,
+  set: (name) => emit('update:name', name),
+});
+const emailProxy = computed({
+  get: () => props.email,
+  set: (email) => emit('update:email', email),
+});
```
basilesimon, line 61: Potential opportunity for [the pattern we discovered this morning](https://github.com/beabee-communityrm/beabee-frontend/pull/134/commits/ea034f7df61b24ed3d362cb469b88b5c0bfabc4b#diff-b3f087794b6db6271d718977b85b52c815c5af7bfffd460496a6481a044cb88aR38-R45)

basilesimon:
I've had a little play and find it really nice. I was able to navigate through it from the keyboard, and wouldn't have noticed the formio layer had I not seen the code.

I'll defer to Design regarding the different states the page can have (logged in/out, guests, anonymous responses, etc.) :)

src/modules/callouts/components/GuestFields.vue
wpf500: But is it a good pattern though?! :smiley:

PR #125, opened 2022-03-04
Notices list page

wpf500:
image

### DIFF
--- a/src/modules/notices/NoticesPage.vue
+++ b/src/modules/notices/NoticesPage.vue
@@ -1,0 +1,66 @@
+<template>
+  <div class="mb-5 flex justify-between border-primary-40 border-b pb-3">
+    <PageTitle :title="t('menu.notices')"></PageTitle>
+    <div class="flex-0 ml-3">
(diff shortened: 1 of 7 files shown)


src/modules/notices/NoticesPage.vue
```diff
@@ -0,0 +1,116 @@
...
+    :headers="headers"
+    :items="noticesTable.items"
+    class="w-full"
+  >
+    <template #status="{ value }">
+      <AppItemStatus :status="value" />
+    </template>
+    <template #name="{ item, value }">
+      <router-link
+        :to="'/notices/' + item.id"
```
basilesimon: I must be missing a bit of context here – we also have `item.url`, which as far as tests are concern do resolve?

basilesimon (commented):
Cool, this is a neat structure!

src/modules/notices/NoticesPage.vue
wpf500: `item.url` is the link you want the notice to point to when a user clicks it
image

/notices/`item.id` is the admin page to edit the notice (currently doesn't exist). I should make this point to the legacy app's page for now, will do that now

PR #126, opened 2022-03-04
fix: checkbox alignment in /join (close #44)

### LINKED ISSUE #44: Fix checkbox and label alignment
Adding here notes on the correct label+checkbox layout after looking at the join form:
- the checkbox should be aligned to the baseline of the label text
- the text shouldn't wrap around the checkbox
- margin between checkbox and text is 4px

Screenshots from the current checkboxes, taken from join and setup pages (the join label is edited to be longer, to show the text wrap);
and one screenshot of the checkbox from the design spec for reference.

checkbox-join
checkbox-setup
checkbox-spec

basilesimon:


wpf500:
:+1:

PR #128, opened 2022-03-04
feat: improve slightly the search box UI

basilesimon:
https://user-images.githubusercontent.com/5701152/156788879-6abdb0e5-2a39-4eb8-9718-2528895e08a1.mov

### DIFF
--- a/src/components/forms/AppSearchInput.vue
+++ b/src/components/forms/AppSearchInput.vue
@@ -4,7 +4,24 @@
     @submit.prevent="emit('update:modelValue', searchText)"
   >
     <AppInput v-model="searchText" class="pr-8" :placeholder="placeholder" />
-    <button class="absolute right-0 w-8 h-full">
+    <button
+      v-if="searchText.length > 0"
+      class="
+        absolute
+        right-5
+        w-8
+        h-full
+        color-red
+        text-primary-70
+        hover:text-primary-40
+      "
+      @click="clearField"
+    >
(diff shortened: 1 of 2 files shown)


src/components/forms/AppSearchInput.vue
```diff
@@ -4,7 +4,24 @@
     @submit.prevent="emit('update:modelValue', searchText)"
   >
     <AppInput v-model="searchText" class="pr-8" :placeholder="placeholder" />
```
wpf500: I'm guessing pr-8 needs to update to pr-16 if the clear button is present, so the input text doesn't overflow into the button

src/components/forms/AppSearchInput.vue
```diff
@@ -22,6 +39,8 @@ const props = defineProps<{
 
 const searchText = ref(props.modelValue);
 
+const clearField = () => (searchText.value = '');
```
wpf500: I think this needs to `emit('update:modelValue', '')` instead of setting `searchText`, as my understanding of the spec is that it should update the search results too

`searchText` is the internal (unsubmitted) value of the search box, triggering an update on `modelValue` is how to propagate the change out of the component.

wpf500 (changes requested):

basilesimon:
@wpf500 thank you very much for your eagle eye!

wpf500 (approved):

PR #134, opened 2022-03-16
Feat: "create a callout" flow

basilesimon:
See [kanban of stories](https://github.com/orgs/beabee-communityrm/projects/1)


basilesimon:
I'm coming to realise that all "steps" include some furniture and boilerplate: a title, navigation UI and logic, emission of events and validations... and that this could be a nice refactor.

The approach I've chosen to navigate so far is hitting a wall: not only do we need to navigate both ways (prev/next), we also can't navigate everywhere at all times. I'm resisting the urge to rely on mutating the main reactive steps store... but would love us to work through this together, Will.

basilesimon:
@wpf500 I'd be keen on taking stock from here. I'm not super happy about how some things look but not quite sure what colours are supposed to be. Let's go through the task list once more when you have time?

src/modules/callouts/pages/CreateCallout.vue
```diff
@@ -0,0 +1,145 @@
...
+};
+export type Steps = Array<Step>;
+
+const steps: Steps = reactive([
+  {
+    id: 'content',
+    name: 'Callout content',
+    description: 'Build your callout with questions and prompts',
+    validated: false,
+    component: markRaw(ContentStep),
```
basilesimon: Presumably because `ContentStep` itself will be updating/changing while the user interacts with it?

src/modules/callouts/components/ContentStep.vue
```diff
@@ -1,22 +1,24 @@
 <template>
-  <AppHeading>Callout content</AppHeading>
-  <div class="grid grid-cols-2 gap-6">
-    <div class="col-span-1">
-      <AppTextArea
-        v-model="dataProxy.introText"
-        label="Introduction text"
-        required
+  <div>
```
basilesimon: Cheers for fixing that no root node error!

src/utils/api/callout.ts
```diff
@@ -53,6 +54,17 @@ export async function fetchCallout(id: string): Promise<GetMoreCalloutData> {
...
+  calloutData: CreateCalloutData
+): Promise<GetBasicCalloutData> {
+  const { data } = await axios.post<Serial<GetBasicCalloutData>>(
+    '/callout',
+    // TODO: passing calloutData directly is not safe, it could contain extra properties
+    calloutData
+  );
+  return toCallout(data);
+}
+
```
basilesimon, line 67: That will do for now!

src/modules/callouts/pages/CreateCallout.vue
wpf500: This just fixes the error in the console about marking a component as reactive. `markRaw` just tells Vue to exclude the component. It's really only the `data` and `validated` properties that need to be reactive as the rest don't change in any case.

> Vue received a Component which was made a reactive object. This can lead to unnecessary performance overhead, and should be avoided by marking the component with markRaw or using shallowRef instead of ref.

src/utils/api/callout.ts
basilesimon: Would love to talk about safety as I'm not 100% clear about the risks, given that the requests appears to require admin auth and [casts the payload into our poll data structure.](https://github.com/beabee-communityrm/beabee/blob/c64684c6c65c33595904cbf9460ca99da577f457/src/api/controllers/CalloutController.ts#L34-L51)

wpf500: It's not a security risk, but the API will reject any request that contains extra properties, e.g.

```ts
// Accepted
{
  "name": "blah"
}

// Rejected
{
  "name": "blah",
  "foo": "bar"
}
```

Unfortunately there isn't a way in TypeScript to say the object must be **just** the properties in `CreateCalloutData`, so `calloutData` could have extra properties and passing it directly to the API means it would then be rejected. We should really rebuild the payload for the request, but that is a lot of boilerplate! Maybe there is a better way

basilesimon: So, as this is not "wired" yet (insofar as `modules/callouts/pages/CreateCallout.vue` does not call `createCallout()` yet), would you see `createCallout()` passing its param `calloutData: CreateCalloutData` through a kind of sanitisation function _before_ posting to the API?

basilesimon:
I've updated [the board of tasks](https://github.com/orgs/beabee-communityrm/projects/1) and this is what remains:

- wiring the frontend to perform the POSTing of the new callout to the API,
- the displaying of the new callout summary,
- a bug in the progress indicator not showing,
- and some questions regarding how this should work on mobile.

src/modules/callouts/pages/CreateCallout.vue
```diff
@@ -138,7 +140,22 @@ const steps: Steps = reactive([
...
+  excerpt: '',
+  access: 'member',
+  allowUpdate: true,
+  allowMultiple: true,
+  image: 'string',
+  starts: new Date(),
+  expires: new Date(),
+});
+
+const submitForm = () => createCallout(makeCalloutData(steps));
```
basilesimon: @wpf500 this is the boilerplate bit I mentioned shamefully. Could also live somewhere else?

src/modules/callouts/pages/CreateCallout.vue
```diff
@@ -2,11 +2,11 @@
   <div class="mb-5 flex justify-between border-primary-40 border-b pb-3">
     <PageTitle :title="t('createCallout.title')" />
   </div>
-  <div class="grid grid-cols-4 gap-8">
-    <div class="col-span-1">
+  <div class="flex gap-8">
+    <div class="flex-0 basis-[240px]">
```
basilesimon: `basis-menu` for consistency with `TheMenu.vue`?

src/components/forms/AppInput.vue
```diff
@@ -40,11 +41,12 @@ import InfoMessage from '../InfoMessage.vue';
 const props = withDefaults(
   defineProps<{
     modelValue?: string;
-    inputType?: 'password' | 'email' | 'text';
+    inputType?: 'password' | 'email' | 'text' | 'date';
```
basilesimon, line 44: Nice, didn't know that!

src/modules/callouts/callouts.route.ts
```diff
@@ -31,7 +31,16 @@ export const calloutsRoute: Array<RouteRecordRaw> = [
...
   },
   {
-    path: '/admin/callouts/:id',
+    path: '/admin/callouts/new',
+    name: 'new-callout',
+    component: () => import('./pages/CreateCallout.vue'),
+    meta: {
+      pageTitle: t('menu.callouts'),
+      role: 'admin',
+    },
```
basilesimon, line 40: 👌

src/modules/callouts/pages/CalloutAdminOverviewPage.vue
```diff
@@ -114,5 +124,8 @@ const props = defineProps<{
 }>();
 const { t } = useI18n();
 
+const route = useRoute();
+const hasCreated = route.query.created !== undefined;
```
basilesimon: How about `wasCreated` or `wasJustCreated`?

basilesimon (commented):
- Not really sure about our terminology, `components/ContentStep.vue` being the final straw for me. I'd refactor into `steps/` then `steps/Step-Something.vue`
- Really practical to have the steps typed properly!

aiscarvalho:
We think it's good to go!

src/components/forms/AppInput.vue
wpf500: I think you made this change! :)

src/modules/callouts/pages/CalloutAdminOverviewPage.vue
wpf500: :+1:

src/modules/callouts/pages/CreateCallout.vue
wpf500: :+1:

PR #143, opened 2022-03-28
Chore/design tweaks

wpf500:
A few tweaks to fix/change 

#### Hide empty payments history

image

#### Fix pagination colours

image
image

#### Fix profile info width

image
image

#### Always put profile info below callouts as it gets lost on the right on big screens

image
image

### DIFF
--- a/src/components/AppPagination.vue
+++ b/src/components/AppPagination.vue
@@ -2,10 +2,8 @@
   <ul class="flex items-center -mx-2">
     <li>
       <button
-        :class="
-          isFirst &&
-          'opacity-25 p-2 bg-primary hover:bg-white hover:font-bold hover:border-link'
-        "
+        class="p-2"
+        :class="isFirst && 'opacity-25 hover:font-bold hover:border-link'"
         :disabled="isFirst"
         @click="emit('update:modelValue', modelValue - 1)"
       >
@@ -16,7 +14,9 @@
       <li v-if="index > 0 && pages[index - 1] !== page - 1">&hellip;</li>
       <li>
         <button
-          :class="page === modelValue && 'text-link bg-white'"
+          :class="
+            page === modelValue && 'text-link bg-white font-bold border-link'
+          "
           :disabled="page === modelValue"
           class="
             leading-none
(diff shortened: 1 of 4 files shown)


src/components/AppPagination.vue
```diff
@@ -2,10 +2,8 @@
   <ul class="flex items-center -mx-2">
     <li>
       <button
-        :class="
-          isFirst &&
-          'opacity-25 p-2 bg-primary hover:bg-white hover:font-bold hover:border-link'
-        "
+        class="p-2"
+        :class="isFirst && 'opacity-25 hover:font-bold hover:border-link'"
```
basilesimon, line 6: Why breaking it down into `class=""` and `:class=""`?

basilesimon, line 6: This commit also makes painfully clear that I've not worked out how to switch between different "themes", so to speak. So what made sense in our salmon-ish theme actually doesn't match designs – or, at least, the mapping isn't clear in my mind.

src/modules/home/components/ContributionInfo.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="flex pt-3 w-76">
+  <div class="flex pt-3 w-[17rem]">
```
basilesimon, line 2: Oh, this is cool! I think we talked about these [arbitrary values](https://tailwindcss.com/docs/width#arbitrary-values) already.

basilesimon (approved):

basilesimon:
@beabee-communityrm/design will let youse review the last item, then I'd like to squash all of these into one commits and merge.

src/components/AppPagination.vue
wpf500: Because `p-2` should always be applied and the rest are conditional, I guess the alternative would be
```
:class="isFirst ? 'p-2 opacity-25 hover:font-bold hover:border-link' : 'p-2'"
```

wpf500: Yeah we should chat about this as I'm also struggling to do the mapping sometimes and end up just guessing.

rlafuente: Let us know if we can be of help here too :hand:

basilesimon: Totes 👊

rlafuente:
Super! No remarks here!

PR #144, opened 2022-03-28
Add internationalisation to create callout flow

wpf500:
https://github.com/beabee-communityrm/beabee-frontend/pull/145 but for the create callout flow

I have defined hopefully a fairly clean structure (if a little long) for the i18n keys, then by defining `inputT` it becomes simple to access them from within each step. There is lots of extra cleanup and simplification that could be done here!

### DIFF
--- a/locales/de.json
+++ b/locales/de.json
@@ -157,26 +157,34 @@
   },
   "createCallout": {
     "steps": {
+      "content": {
+        "description": "",
+        "inputs": {
+          "intro": {
+            "help": "",
+            "label": "",
+            "placeholder": ""
+          }
+        },
+        "title": ""
+      },
       "titleAndImage": {
         "description": "",
         "inputs": {
           "description": {
-            "title": ""
+            "help": "",
+            "label": "",
+            "placeholder": ""
           },
           "image": {
             "help": "",
-            "placeholder": "",
-            "title": ""
-          },
-          "intro": {
-            "help": "",
-            "placeholder": "",
-            "title": ""
+            "label": "",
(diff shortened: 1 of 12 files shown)


basilesimon (commented):
Really interesting, this library. I would have thought there'd be a more vue-esque solution but quite pleased that this is so general-purpose 🙌

The one thing I'd raise is your choice to selectively enable markdown parsing. What's the case against turning it on globally, overwriting what `t()` does for us throughout?

wpf500:
I originally did this using [vue-markdown-render](https://github.com/cloudacy/vue-markdown-render) but then thought why am I adding 20kb to the client payload when it can just be precompiled.

A few different reasons for the opt-in:
* Not accidentally causing formatting issues in old strings
* The markdown renderer wraps the string in `<p></p>` tags which wouldn't be appropriate in most cases (although I'm sure this can be disabled)
* An idea that at some point in the future the opt in could be linked to the codebase so only things marked with `:md` can be injected via `v-html`, eliminating lots of possible arbitrary HTML injection. Although having said that arguably it would be much better to run everything through the markdown renderer as it would sanitise any HTML!

basilesimon:
To me, the risk of forgetting whether a Google Sheet cell has markdown parsing is greater than the potential rogue `*` we might end up with if pushing markdown-parsing-by-default in the whole app.

But with that said, I'm happy to defer to your judgment and will let you squash and merge or articulate it differently given your other pull request.

wpf500:
As discussed on a call I agree it would be better to have markdown by default but am merging this as is for now

PR #146, opened 2022-03-28
feat: add callouts admin list page

wpf500:
One change from spec:
* Eye icon is `text-body-40` when the callout is not hidden (rather than when hidden), I think we should be highlighting when a callout is hidden not the other way round.

A couple of things deferred for simplicity that I will add to the backlog:
* Count next to each status (open, scheduled, etc.)
* Number of responses per callout

image

### DIFF
--- a/src/modules/callouts/callouts.route.ts
+++ b/src/modules/callouts/callouts.route.ts
@@ -21,4 +21,13 @@ export const calloutsRoute: Array<RouteRecordRaw> = [
       noAuth: true,
     },
   },
+  {
+    path: '/admin/callouts',
+    name: 'calloutsAdmin',
+    component: () => import('./pages/CalloutsAdminPage.vue'),
+    meta: {
+      pageTitle: t('menu.callouts'),
+      role: 'admin',
+    },
+  },
 ];
--- a/src/modules/callouts/pages/CalloutsAdminPage.vue
+++ b/src/modules/callouts/pages/CalloutsAdminPage.vue
@@ -41,7 +41,7 @@
         </template>
         <template #hidden="{ value }">
           <font-awesome-icon
-            :class="!value && 'text-body-40'"
+            :class="value ? 'text-body-80' : 'text-body-40'"
             :icon="value ? 'eye-slash' : 'eye'"
           />
         </template>
--- a/src/modules/callouts/pages/CalloutsAdminPage.vue
+++ b/src/modules/callouts/pages/CalloutsAdminPage.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="mb-5 flex justify-between border-primary-40 border-b pb-3">
     <PageTitle :title="t('menu.callouts')"></PageTitle>
-    <div class="flex-1 md:hidden">
+    <div class="flex-1 block lg:hidden">
       <AppSelect v-model="currentStatus" :items="statusItems" />
     </div>
     <div class="flex-0 ml-3">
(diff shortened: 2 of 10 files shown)


src/modules/callouts/callouts.route.ts
```diff
@@ -21,4 +21,13 @@ export const calloutsRoute: Array<RouteRecordRaw> = [
       noAuth: true,
     },
   },
+  {
+    path: '/admin/callouts',
```
basilesimon, line 25: We don't have anything else at `/admin`, correct?

src/modules/callouts/pages/CalloutsAdminPage.vue
```diff
@@ -0,0 +1,229 @@
+<template>
+  <div class="mb-5 flex justify-between border-primary-40 border-b pb-3">
+    <PageTitle :title="t('menu.callouts')"></PageTitle>
+    <div class="flex-1 md:hidden">
+      <AppSelect v-model="currentStatus" :items="statusItems" />
+    </div>
+    <div class="flex-0 ml-3">
+      <AppButton href="/tools/polls">{{
```
basilesimon, line 8: `href="/tools/polls"` 

is this right?

src/modules/callouts/pages/CalloutsAdminPage.vue
```diff
@@ -0,0 +1,229 @@
...
+        t('calloutsAdmin.addCallout')
+      }}</AppButton>
+    </div>
+  </div>
+  <div class="md:flex">
```
basilesimon, line 13: The `flex` here is a little bit out whack from ~1,080px down. 

Now, I remember "desktop-first for admins", but just checking with @beabee-communityrm/design that this needs fixing indeed.

aiscarvalho:
It looks and works very well!

We pushed some minor adjustments to the text and were wondering about the following:
- could the dates be displayed with the month truncated to the first three letters?  
They look too wide at the moment. 
- could the currently active sorting column title have `font-bold`, color `text-primary` and a right aligned caret?
image
- (this one probably fixed in a different branch!) there a background color in the inactive pagination arrows 
image

src/modules/callouts/callouts.route.ts
wpf500: No but the create callout flow needs to move to `/admin/callouts/:id` and probably the notices list page to `/admin/notices`

src/modules/callouts/pages/CalloutsAdminPage.vue
wpf500: It is for now until the create callout flow is ready

src/modules/callouts/pages/CalloutsAdminPage.vue
aiscarvalho: The filters on the left become a dropdown below 768px. They could switch to dropdown at <1080px to make more space for the table. Do you think that would work?

basilesimon: image

Looks a bit clunky before the title becomes a banner, like under 768px. But feels more readable!

wpf500:
@aiscarvalho I have implemented the sorting column changes as suggested but I don't think the caret should be right aligned, on wide columns it gets lost:
image

I'm going to merge this PR with left aligned carets as I need to merge this to get going on the next piece but I can create another PR If you disagree with that decision

rlafuente:
This works -- issue done :ok_hand:

src/modules/callouts/pages/CalloutsAdminPage.vue
aiscarvalho: Absolutely!
This is what I had in mind and didn't manage to explain:
image

Would that be a quick fix the ~1080 whack?

PR #148, opened 2022-03-30
feat: refactor nav and add admin menu

wpf500:
Implementation of the admin nav. I have used this as an opportunity to refactor the menu code:

* Remove magic 240px by adding it as a `menu` spacing option to Tailwind
* Reworked how styles are applied so there is no custom CSS
* Separated sections out into into separate components

image

### DIFF
--- a/src/layouts/menu/TheMenuList.vue
+++ b/src/layouts/menu/TheMenuList.vue
@@ -1,36 +1,32 @@
 <template>
   <div class="flex-auto">
-    <nav
-      v-for="(section, index) in sections"
+    <TheMenuListSection
+      v-for="(section, index) in menu"
       :key="index"
-      class="menu-section"
-      :class="{ 'is-settings': section.type === 'settings' }"
-    >
-      <div v-if="index !== 0" class="menu-title">
-        {{ section.title }}
+      :section="section"
+      :is-first="index === 0"
+    />
+    <div v-if="canAdmin" class="bg-primary-10 py-4 mt-4">
+      <TheMenuListSection
+        v-for="(section, index) in adminMenu"
+        :key="index"
+        :section="section"
+        :is-first="index === 0"
+      />
+      <div class="mx-4">
+        <a href="/members">
+          <TheMenuListItem
+            :icon="['fa', 'cogs']"
+            :title="t('menu.adminDashboard')"
+          />
+        </a>
       </div>
-      <ul class="flex flex-col">
-        <template v-for="(item, itemIndex) in section.items" :key="itemIndex">
-          <li v-if="!item.role || currentUserCan(item.role).value">
-            <router-link class="menu-item" :to="item.href">
-              <TheMenuListItem :icon="item.icon" :title="item.title" />
-            </router-link>
-          </li>
-        </template>
-      </ul>
-    </nav>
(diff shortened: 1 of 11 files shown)


rlafuente (approved):
This is great!

src/layouts/menu/TheMenuList.vue
```diff
@@ -42,56 +38,20 @@
...
+import { canAdmin } from '../../utils/currentUserCan';
 import { logout } from '../../utils/api/auth';
 
-const { t } = useI18n();
+import { menu, adminMenu } from './menu-list';
 
-defineProps<{
-  sections: MenuSection[];
-}>();
+adminMenu;
```
basilesimon: Sorry, I don't know why this is needed?

basilesimon (changes requested):

src/layouts/menu/TheMenuListItem.vue
```diff
@@ -1,16 +1,46 @@
...
+          'text-primary-80 hover:bg-primary-5',
+          'text-primary-80 bg-primary-20',
+        ],
+        settings: [
+          'text-grey-dark hover:bg-grey-lighter',
+          'bg-primary-5 text-grey-darker',
+        ],
+      } as const
+    )[props.type][props.isActive ? 1 : 0])
+);
```
basilesimon: 🥲

src/layouts/menu/menu-list.ts
```diff
@@ -44,4 +40,31 @@ export default [
...
+      },
+      {
+        title: t('menu.notices'),
+        href: '/admin/notices',
+        icon: ['fa', 'bullhorn'],
+        isActive: /^\/admin\/notices.*/,
+      },
+    ],
+  },
+];
```
basilesimon, line 70: Also, and kind of related to another thread on the admin dashboard:

I would really see the dashboard being at the top of the list, and potentially either the default page or at the root of the route (geddit?)

src/layouts/menu/TheMenuList.vue
wpf500: Oops will remove, thanks!

src/layouts/menu/menu-list.ts
wpf500: Yeah I agree on all points. The "Dashboard" I have in the menu at the moment is actually just a link to the legacy app (as some functionality still only exists there). It needs a better name so we can add the actual dashboard to the menu!

src/layouts/menu/TheMenuListItem.vue
wpf500: This is very strange code, could it be clearer somehow? Maybe defining the object outside the `computed`?

```ts
const itemClasses = {
  main: [
    'text-primary-80 hover:bg-primary-5',
    'text-primary-80 bg-primary-20',
  ],
  settings: [
    'text-grey-dark hover:bg-grey-lighter',
    'bg-primary-5 text-grey-darker',
  ],
} as const;

const blah = computed(() => itemClasses[props.type][props.isActive ? 1 : 0]);
```

basilesimon: 👍

basilesimon (approved):

PR #149, opened 2022-03-30
Callout admin summary page

wpf500:
image

### DIFF
--- a/src/modules/callouts/callouts.route.ts
+++ b/src/modules/callouts/callouts.route.ts
@@ -30,4 +30,33 @@ export const calloutsRoute: Array<RouteRecordRaw> = [
       role: 'admin',
     },
   },
+  {
+    path: '/admin/callouts/:id',
+    name: 'calloutAdmin',
+    component: () => import('./pages/CalloutAdminPage.vue'),
+    meta: {
+      pageTitle: t('menu.callouts'),
+      role: 'admin',
+    },
+    children: [
+      {
+        path: '',
+        name: 'calloutAdminOverview',
+        component: () => import('./pages/CalloutAdminOverviewPage.vue'),
+        meta: {
+          pageTitle: t('menu.callouts'),
+          role: 'admin',
+        },
+      },
+      {
+        path: 'responses',
+        name: 'calloutAdminResponses',
+        component: () => import('./pages/CalloutAdminResponsesPage.vue'),
+        meta: {
+          pageTitle: t('menu.callouts'),
+          role: 'admin',
+        },
+      },
+    ],
+  },
 ];
--- a/src/modules/callouts/pages/CalloutAdminOverviewPage.vue
+++ b/src/modules/callouts/pages/CalloutAdminOverviewPage.vue
@@ -33,20 +33,33 @@
       <AppHeading>{{ t('calloutAdminOverview.settings.label') }}</AppHeading>
       <AppInfoList>
         <AppInfoListItem
-          :name="t('calloutAdminOverview.settings.openTo')"
-          value="Members only"
+          :name="t('calloutAdminOverview.settings.openTo.label')"
+          :value="
+            callout.access === 'member'
+              ? t('calloutAdminOverview.settings.openTo.membersOnly')
+              : t('calloutAdminOverview.settings.openTo.everyone')
+          "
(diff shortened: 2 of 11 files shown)


wpf500:
I've made some progress on the summary page but have a question for you @beabee-communityrm/design.

I'd like your opinion on the position of the action buttons. Currently on large monitors they get a bit lost on the right, on smaller devices I've made them go above the rest of the page (see screenshots). I think the position of the buttons on smaller devices would work better for all screen widths. What do you think?

In general I think we could do with coming up with a clearly defined grid. Currently lots of our pages use different grids and so have different column widths etc and some pages look strange on large monitors because we have no defined max width. Can we schedule a call about this?

image

image

rlafuente:
Cheers!

Using the mobile layout for the action buttons is a good way to unify the looks and solve the problem, but we end up missing out on the ample width of desktops. What do you think about keeping the buttons on a left-aligned column next to the main content column? Like this:

image

We'll follow up on Slack regarding the call

wpf500:
Happy to go with this method but we need to define what the maximum width for the column is, let's discuss later!

src/modules/callouts/pages/CalloutsAdminPage.vue
```diff
@@ -32,12 +32,12 @@
           <AppItemStatus :status="value" />
         </template>
         <template #title="{ item, value }">
-          <a
-            :href="'/tools/polls/' + item.slug"
+          <router-link
+            :to="'/admin/callouts/' + item.slug"
```
basilesimon, line 36: Also needs to change [here](https://github.com/beabee-communityrm/beabee-frontend/blob/94239ec714f16b77026da310a00929af1d89d5ff/src/modules/callouts/pages/CalloutsAdminPage.vue#L8-L10)

src/modules/callouts/callouts.route.ts
```diff
@@ -30,4 +30,33 @@ export const calloutsRoute: Array<RouteRecordRaw> = [
       role: 'admin',
     },
   },
+  {
+    path: '/admin/callouts/:id',
+    name: 'calloutAdmin',
+    component: () => import('./pages/CalloutAdminPage.vue'),
```
basilesimon, line 36: Just on the topic of _naming things_, and from my sleep-deprived perspective:

Could this routing between be an opportunity to name pages differently, e.g.

* `/admin/callouts        =>   ./pages/TheCalloutsAdminPage.vue`
* `/admin/callouts/:id   =>   ./pages/CalloutAdminPage.vue or EditCalloutPage.vue`

src/modules/callouts/pages/CalloutAdminOverviewPage.vue
```diff
@@ -0,0 +1,118 @@
...
+              {{ formatDistanceLocale(new Date(), callout.starts) }})</span
+            >
+          </span>
+          <span v-else>–</span>
+        </AppInfoListItem>
+        <AppInfoListItem
+          :name="t('calloutAdminOverview.dates.ends')"
+          :value="callout.expires && formatLocale(callout.expires, 'PPP')"
+        />
+      </AppInfoList>
```
basilesimon, line 39: Sweet

src/modules/callouts/pages/CalloutAdminResponsesPage.vue
```diff
@@ -0,0 +1 @@
+<template>here</template>
```
basilesimon, line 1: Strong work

basilesimon:
Amazing, thanks for putting this together so quickly 🙌

I added a couple of comments above but found nothing jarring. The only thing I'd like to emphasise is the naming which, while it makes sense, takes a second to get used to. If I understand this right, it's something like this:

```
/callouts                     → CalloutsPage (list of callouts, for users)
/callouts/:id                 → CalloutPage (single callout, for users)

/admin/callouts                → CalloutsAdminPage
/admin/callouts/:id            → CalloutAdminPage
                 |→ /          → CalloutAdminOverviewPage
                 |→ /responses → CalloutAdminResponsesPage
```

Which brings us back to something we discussed early on: how we've organised things on the filesystem v. on the router.

rlafuente (approved):
Super!

PR #150, opened 2022-03-31
feat: add max width for dashboard

wpf500:


aiscarvalho:
@wpf500 the max-width works fine and we can fine-tune the value later if needed!
We noticed that user Settings (Your information and Contribution pages) don't get the max-width applied to them.

wpf500:
Ah I fixed this @aiscarvalho but didn't update dev. Merging :+1:

PR #155, opened 2022-04-06
Feat/some tweaks

wpf500:


src/modules/home/HomePage.vue
```diff
@@ -37,9 +37,13 @@
       t('homePage.openCallouts')
     }}</AppHeading>
 
-    <div class="flex mb-4">
-      <!-- just show the first callout for now (design decision) -->
-      <CalloutCard :callout="callouts[0]" />
+    <div class="flex -mx-3 my-6 flex-wrap">
+      <CalloutCard
+        v-for="callout in callouts"
```
basilesimon, line 42: Should we limit how many max?

wpf500: I've added a limit of 3

basilesimon:
I like it!

PR #157, opened 2022-04-06
Proposal: using a `title` font

### LINKED ISSUE #147: Support for several font-families
Presently, `<body>` is set to follow `.body-font` (Open Sans by default).

While it is possible to specify additional fonts in `theme.json` (it's even implemented in `tailwind.config.js`, these declarations are not used in the app.

That behaviour feels unexpected to me, and was reported by Sascha on Slack.

[Here's an example reproduction of the behaviour](https://github.com/beabee-communityrm/beabee-frontend/commit/af725021b8301fbef89894456b3e7d46e6f0db01)

### Possible action routes
@beabee-communityrm/design would need to flag which UI elements follow `.title-font`, and the rest will follow `.body-font`.

basilesimon:


rlafuente:
This is fab! :tada: And we think it's good to go, with the sidenote that Roboto titles should be reverted to Open Sans before merging.

(we can definitely review the designs and default typography, but we should keep that for its own ticket/issue/backlog item)

basilesimon:
@wpf500 Marking you as reviewer as an FYI. If you're happy I'd like to ship this to both `dev` and `main` – as it's using Open Sans throughout, it should not be a visible change. But Sascha will be happy :)

saschafoerster:
> @wpf500 Marking you as reviewer as an FYI. If you're happy I'd like to ship this to both `dev` and `main` – as it's using Open Sans throughout, it should not be a visible change. But Sascha will be happy :)

😎

wpf500 (approved):

wpf500:
Well if it makes @saschafoerster happy :)

saschafoerster:
https://github.com/beabee-communityrm/beabee-frontend/issues/147#issuecomment-1105788369

Issue #158, opened 2022-04-07
Wireframe: post-signup dashboard


aiscarvalho:
A wireframe for the dashboard after signup with suggested actions and necessary actions ([penpot link](https://design.penpot.app/#/view/167be840-9008-11ec-bd38-efdb6fa63305?page-id=4fe6fd30-b4f8-11ec-a6b4-8bf5c7f95f90&section=interactions&index=0&share-id=592877f0-b5c2-11ec-862a-da95c03a630b)):

#### Dashboard:
![image](https://user-images.githubusercontent.com/122730/162251295-0e4aae05-7986-4bca-a2e2-76bf6c341a33.png)

#### Dashboard with notice (and completed or dismissed suggested actions)
![image](https://user-images.githubusercontent.com/122730/162251270-6cae1220-fb12-4ef3-92bb-e4c3d029d949.png)

We hope we've captured everything, let us know if there's something that could be added/changed :)

aiscarvalho:
Today we had a great call and super useful feedback from Julia and Tobias. Here's a quick list of the updates and the revised wireframes:

- post-signup suggestions should all be _dismissable_. Options to provide: don't show this again; remind me later (shows in next login)
- after signup we'll use warnings to prompt admin for missing data/actions
- payment copy box: make it re-purposable. It's the place to show information regarding payment according to each newsroom situation. Some examples:
    - a default copy "we won't charge you yet"
    - you're paying X/month, next payment is in *Y of Mon of year*
    - you're a tester and won't be charged
    - include link to "edit payment details"
- remove "subscription" and use "new members"
- add illustrations to import contacts / create join form
- add estimated time to import contacts / create join form
- move support up and include Tobias portrait
- move "join slack" link down

![image](https://user-images.githubusercontent.com/122730/163005446-88d5885f-b611-4ed6-a441-ef1885e7e49a.png)

---
![image](https://user-images.githubusercontent.com/122730/163005662-02692cf7-1dd4-4276-b31b-b07e7d90ab08.png)

wpf500:
I think this looks good for now. I worry a bit about the whole thing being dismissable, but maybe the notices will be able to cover the important stuff. I think this will be a lot clearer when we start implementing it and get a feel for it, so :+1:

TobiasHauswurz:
Yeah, looks good for now. The thing about being dismissable: 

[...]

PR #160, opened 2022-04-13
feat: dashboard notice for when there's no latest callout

rlafuente:
A first stab at the notice for when there's no active callout. The code presumes that `latestCallout` would be empty in the case there's no active callouts. Is this right?

(Not to be merged until we decide on a proper design for this element and implement it!)

Quick preview:

image

### DIFF
--- a/src/modules/admin/AdminPage.vue
+++ b/src/modules/admin/AdminPage.vue
@@ -45,14 +45,16 @@
       </div>
     </div>
     <div class="flex-1 basis-2/3">
-      <AppHeading>{{ t('adminDashboard.latestCallout.title') }}</AppHeading>
-      <p v-if="latestCallout" class="block bg-white p-4 mt-4 mb-8 rounded">
-        {{ t('adminDashboard.latestCallout.empty') }}
-        <router-link to="/admin/callouts/new" class="text-link">
-          {{ t('adminDashboard.latestCallout.createNew') }}
-        </router-link>
-      </p>
-      <CalloutSummary v-else-if="latestCallout === null" :callout="latestCallout" footer />
+      <div class="block bg-white p-4 mt-4 mb-8 rounded">
+        <AppHeading>{{ t('adminDashboard.latestCallout.title') }}</AppHeading>
+        <CalloutSummary v-if="latestCallout" :callout="latestCallout" footer />
+        <div v-else-if="latestCallout === null">
+          {{ t('adminDashboard.latestCallout.empty') }}
+          <router-link to="/admin/callouts/new" class="text-link">
+            {{ t('adminDashboard.latestCallout.createNew') }}
+          </router-link>
+        </div>
+      </div>
       <div
         class="
           p-10
--- a/src/modules/admin/AdminPage.vue
+++ b/src/modules/admin/AdminPage.vue
@@ -51,6 +51,15 @@
           <CalloutSummary :callout="latestCallout" footer />
         </a>
       </template>
+      <template v-else>
+        <AppHeading>{{ t('adminDashboard.latestCallout.title') }}</AppHeading>
+        <p class="block bg-white p-4 mt-4 mb-8 rounded">
+          No currently active callouts.
+          <router-link to="/admin/callouts/new" class="text-link">
+            Create a new one?
+          </router-link>
+        </p>
+      </template>
       <div
         class="
           p-10
--- a/src/modules/admin/AdminPage.vue
+++ b/src/modules/admin/AdminPage.vue
@@ -45,21 +45,14 @@
       </div>
     </div>
     <div class="flex-1 basis-2/3">
-      <template v-if="latestCallout">
-        <AppHeading>{{ t('adminDashboard.latestCallout.title') }}</AppHeading>
-        <a class="block bg-white p-4 mt-4 mb-8 rounded">
-          <CalloutSummary :callout="latestCallout" footer />
-        </a>
-      </template>
-      <template v-else>
-        <AppHeading>{{ t('adminDashboard.latestCallout.title') }}</AppHeading>
-        <p class="block bg-white p-4 mt-4 mb-8 rounded">
-          No currently active callouts.
-          <router-link to="/admin/callouts/new" class="text-link">
-            Create a new one?
-          </router-link>
-        </p>
-      </template>
+      <AppHeading>{{ t('adminDashboard.latestCallout.title') }}</AppHeading>
+      <p v-if="latestCallout" class="block bg-white p-4 mt-4 mb-8 rounded">
+        {{ t('adminDashboard.latestCallout.empty') }}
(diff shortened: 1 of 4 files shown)


julia-hildebrand:
Looks like a nice simple solution. Couple of questions: 
- Is this the most dominant place where we need to have the view when there are no call-outs? 
- Does it make sense for actions like these to also work wiht buttons to make it more prominant or even different font size.
- Would it be good to think about context of guides here too? 
- As discussed with wireframes, I think the areas in the dashboard should clearly be separated out. Join the beabee slcak community shouldn't be mixed into the area of reviewing current activities, even if there is space. I think we should clearly define the areas and what they do for our user. joining the community is just an additional information that should come last.

src/modules/admin/AdminPage.vue
```diff
@@ -51,6 +51,15 @@
           <CalloutSummary :callout="latestCallout" footer />
         </a>
       </template>
+      <template v-else>
```
wpf500: I think it's probably worth rearranging this a bit so there is less duplication
```vue
<AppHeading... />
<div class="bg-white p-4 mt-4 mb-8 rounded">
  <CalloutSummary v-if="latestCallout".../>
  <p v-else>
    No current callouts etc.
  </p>
</div>
```

src/modules/admin/AdminPage.vue
```diff
@@ -51,6 +51,15 @@
...
+        <AppHeading>{{ t('adminDashboard.latestCallout.title') }}</AppHeading>
+        <p class="block bg-white p-4 mt-4 mb-8 rounded">
+          No currently active callouts.
```
wpf500: Needs to be localised

src/modules/admin/AdminPage.vue
wpf500: Only other thing to consider is that if there is a callout it will show the "No current callouts..." box until the API responds (as `latestCallout` will be `undefined`).

We might considering differentiating between loading and none by allowing `latestCallout` to be `null`, e.g.

```ts

// Add "| null" to the type
const latestCallout = ref<GetBasicCalloutData | null>();

// ...

onBeforeMount(() => {
  // ... other API callouts


  fetchCallouts({
    limit: 1,
    sort: 'starts',
    order: 'DESC',
  }).then((results) => {
   // Assign null if no callout
   latestCallout.value = results.count > 0 ? results.items[0] : null;
  });
});
```

Then the `v-else` in the template becomes `v-else-if="latestCallout === null"`

wpf500 (changes requested):

wpf500:
@julia-hildebrand 
> Looks like a nice simple solution. Couple of questions:
> * Is this the most dominant place where we need to have the view when there are no call-outs?

Not sure what this means, could you clarify?

> * Does it make sense for actions like these to also work wiht buttons to make it more prominant or even different font size.

I think the simple text approach works well on the dashboard

> * Would it be good to think about context of guides here too?
> * As discussed with wireframes, I think the areas in the dashboard should clearly be separated out. Join the beabee slcak community shouldn't be mixed into the area of reviewing current activities, even if there is space. I think we should clearly define the areas and what they do for our user. joining the community is just an additional information that should come last.

Can we address these in a separate issue as it isn't strictly related to the no callout notice?

rlafuente:
Thanks Julia for all the good points -- we were just picking this up since it was a low-hanging fruit to begin diving into the frontend tasks. We can hold this until the dashboard designs are settled, but we already [updated the wireframes](https://github.com/beabee-communityrm/beabee-frontend/issues/158#issuecomment-1096919383) according to your previous feedback.

julia-hildebrand:
@wpf500 

> > Looks like a nice simple solution. Couple of questions:
> > 
> > * Is this the most dominant place where we need to have the view when there are no call-outs?
> Not sure what this means, could you clarify?
> Was just wondering if there are more places than the dashboard we wanted to look at for when there is no call-out yet. Like the Callout page itself. 
> 
> > * Does it make sense for actions like these to also work with buttons to make it more prominant or even different font size.
> 
> I think the simple text approach works well on the dashboard
> 

> > * Would it be good to think about context of guides here too?
> > * As discussed with wireframes, I think the areas in the dashboard should clearly be separated out. Join the beabee slcak community shouldn't be mixed into the area of reviewing current activities, even if there is space. I think we should clearly define the areas and what they do for our user. joining the community is just an additional information that should come last.
> 
> Can we address these in a separate issue as it isn't strictly related to the no callout notice?

> On guides: Put that here because when there is no call-out it might be because call-outs haven't been used yet at all. That would be a reason to lead user into a guide of how to create a call-out/why when there is nothing there yet. 
 > On slack: Put the comment here because I thought it might have been moved to that place so that it doesn't look empty when there are no callouts showing.

src/modules/admin/AdminPage.vue
rlafuente: Hup, done, added a couple of new entries. Should I mark somewhere when there's new locale strings added?

src/modules/admin/AdminPage.vue
rlafuente: Precious advice, this now looks much cleaner :broom:

src/modules/admin/AdminPage.vue
wpf500: I usually just prod Tobi on Slack

src/modules/admin/AdminPage.vue
```diff
@@ -45,12 +45,14 @@
...
     </div>
     <div class="flex-1 basis-2/3">
-      <template v-if="latestCallout">
-        <AppHeading>{{ t('adminDashboard.latestCallout.title') }}</AppHeading>
-        <a class="block bg-white p-4 mt-4 mb-8 rounded">
-          <CalloutSummary :callout="latestCallout" footer />
-        </a>
-      </template>
+      <AppHeading>{{ t('adminDashboard.latestCallout.title') }}</AppHeading>
+      <p v-if="latestCallout" class="block bg-white p-4 mt-4 mb-8 rounded">
```
wpf500: Looks great but I think the `v-if` and `v-else-if` might be the wrong way round :stuck_out_tongue:

wpf500 (changes requested):

src/modules/admin/AdminPage.vue
wpf500: Oh also I think `<CalloutSummary>` needs to be within the element with the `bg-white p-4...` styles (both cases want that background)

You'll need to make the `<p>` a `<div>` as paragraph tags can't have block elements in them

wpf500 (changes requested):

src/modules/admin/AdminPage.vue
rlafuente: D'oh, of course. All fixed, thank you for the careful review :stars:

wpf500 (approved):

wpf500:
Looks great, I will fix the conflicts and merge

PR #172, opened 2022-04-27
Edit callout

basilesimon:
- New route: `/admin/callours/edit/:id`
- Uses the same `CreateCallout.vue` page, whose role is:
    - Check if a callout with this `:id` exists,
    - Build the `Steps` data with that callout data if it does, and placeholder empties if it doesn't,
    - And then render the callout-building form, as well as dispatching the submit to the API.

Most of the code represents these almost symmetrical mapping functions:

```ts
makeCalloutData = Steps => CreateCalloutData;
makeStepsData = CreateCalloutData => Steps;
``` 

### To do and check
1. PATCH-ing endpoint on the API side
2. Deactivate editing of the slug


basilesimon:
Commits above this comment deal with routes. We promoted a `/admin/callouts/view/:id` in lieu of `edit`, which will now be used to (🥁.......) edit a callout!

basilesimon:
Will, sanity check of https://github.com/beabee-communityrm/beabee-frontend/pull/172/commits/4127876c780319b8f77b9cd03db396f60aa9ff5b please.

This is a simple refactor into a function, and sets the stage for the second rather manual mapping, this `makeStepsData` (which doesn't pass the type checker yet).

basilesimon:
@wpf500 would welcome your thoughts on how to disable editing of the slug.

src/utils/api/callout.ts
```diff
@@ -65,6 +65,11 @@ export async function createCallout(
   return toCallout(data);
 }
 
+export async function editCallout(calloutData: any): Promise<any> {
```
wpf500: I think this should be called `updateCallout` to use the same terminology as other API methods

src/modules/callouts/pages/CreateCallout.vue
```diff
@@ -1,117 +1,149 @@
...
   <div class="mb-5 flex justify-between border-primary-40 border-b pb-3">
-    <PageTitle :title="t('createCallout.title')" />
+    <PageTitle v-if="mode === 'new'" :title="t('createCallout.title')" />
+    <PageTitle
+      v-if="mode === 'edit'"
+      :title="t('editCallout.title')"
+      subTitle="Warning!"
+    />
   </div>
+  <WarningBanner v-if="mode === 'edit'" />
```
wpf500: This should only show on the content step I think as it's only editing the form that's the problem

src/modules/callouts/pages/CreateCallout.vue
```diff
@@ -1,117 +1,149 @@
...
+import WarningBanner from '../components/WarningBanner.vue';
 
 const { t } = useI18n();
-
 const router = useRouter();
+const route = useRoute();
 
-const steps: Steps = reactive({
+const steps: Ref<null | Steps> = ref(null);
+const mode = route.path.split('/')[3];
```
wpf500: :smile:

I'm pretty sure you can use something like `route.name === 'edit-callout'`, i.e. the name in the `callouts.route.ts` file is accessible somehow

src/modules/callouts/pages/CreateCallout.vue
```diff
@@ -144,77 +176,25 @@ const steps: Steps = reactive({
...
+    const newCallout = await createCallout(callout);
+  }
   router.push({
-    path: '/admin/callouts/edit/' + newCallout.slug,
+    path: '/admin/callouts/edit/' + callout.slug,
     query: { created: null },
   });
 }
+onBeforeMount(async () => {
+  const calloutId = route.params.id as string;
```
wpf500, line 121: This method should use either the `async` pattern or `Promise.then().catch()` (I'd prefer `async`/`await`)

I don't think it should default to an empty form if the callout doesn't exist so I wouldn't bother catching the error. Ideally it would go to a 404 page but we don't actually have one of those at the moment :cold_sweat: 

The code could be simplified to something like:
```ts
steps.value = mode === 'edit'
  ? makeStepsData(await fetchCallout(calloutId))
  : makeStepsData({})
```

wpf500: I'm surprised that `{}` passes the type checker, but it would probably be more explicit if `makeStepsData` could accept no parameter (i.e. make the parameter optional)

src/modules/callouts/components/CreateCalloutForm.vue
```diff
@@ -0,0 +1,84 @@
...
+import { useI18n } from 'vue-i18n';
+
+import AppHeading from '../../../components/AppHeading.vue';
+import AppButton from '../../../components/forms/AppButton.vue';
+import Stepper from '../components/Stepper.vue';
+import { Steps } from '../create-callout.interface';
+
+const { t } = useI18n();
+const props = defineProps<{ modelValue: Steps }>();
+const emit = defineEmits(['update:modelValue', 'submit:modelValue']);
```
wpf500: I don't think this needs `update:modelValue`? Also could `submit:modelValue` just be `submit` for consistency?

src/modules/information/InformationPage.vue
```diff
@@ -1,8 +1,5 @@
 <template>
-  <PageTitle
-    :title="t('informationPage.title')"
-    :sub-title="t('informationPage.subTitle')"
-  />
+  <PageTitle :title="t('informationPage.title')" />
```
wpf500: Not sure what this change is for?

wpf500 (changes requested):

wpf500:
> @wpf500 would welcome your thoughts on how to disable editing of the slug.

I think the simplest way would be to pass the `mode` down to the steps so they can modify their behaviour. This would work both for the read only slug and for the warning on the content step.

So something like
```
<CreateCalloutForm :mode="mode">
...
<component :is="selectedStep.component" :mode="mode">
...
```

wpf500:
Also can you rename `CreateCalloutForm` to just `CalloutForm` or another name to reflect it's for both creating and editing

wpf500 (approved):

PR #189, opened 2022-05-12
feat: add empty slot to AppTable, WIP contact list messages

rlafuente:
Very much work-in-progress ~~and not working yet~~, but I need some feedback on implementation details and a PR seems a good way to track those.

~~The main issue I need guidance with: when using search to filter the contact list, cases where there's no results should pop a specific empty message ("No results found"). We also want to accomodate a different empty message for when there's no contacts present at all.~~

~~However, I can't find a variable or state to make a condition against, to determine whether I'm running into an empty state because of a search with 0 results, or because there are no contacts at all. I just need that tidbit to be able to set a proper condition and take care of the other chores (e.g. mark strings as translatable)~~
 
~~(There already is a "no results found" message on the items count area, but it'd be preferable for it to appear in the table itself)~~

Decided to check whether a search query is present in the URL, it seems do to the trick.

### DIFF
--- a/src/modules/contacts/ContactsPage.vue
+++ b/src/modules/contacts/ContactsPage.vue
@@ -19,11 +19,8 @@
         :items="contactsTable.items"
         class="w-full mt-2 whitespace-nowrap"
       >
-        <template #empty v-if="route.query.s">
-          <p>{{ t('contacts.noResults') }}</p>
-        </template>
-        <template #empty v-else>
-          <p>{{ t('contacts.noContacts') }}</p>
+        <template #empty>
+          <p>{{ route.query.s ? t('contacts.noResults') : t('contacts.noContacts') }}</p>
         </template>
         <template #firstname="{ item }">
           <router-link
--- a/src/modules/contacts/ContactsPage.vue
+++ b/src/modules/contacts/ContactsPage.vue
@@ -19,6 +19,12 @@
         :items="contactsTable.items"
         class="w-full mt-2 whitespace-nowrap"
       >
+        <template #empty v-if="contactsTotal && contactsTotal.value">
+          <p>No results found.</p>
+        </template>
+        <template #empty v-else>
+          <p>No contacts added yet.</p>
+        </template>
         <template #firstname="{ item }">
           <router-link
             :to="'/admin/contacts/' + item.id"
@@ -67,7 +73,7 @@
               ><b>{{ n(contactsTable.total) }}</b></template
             >
           </i18n-t>
-          <i18n-t v-else keypath="contacts.noResults" />
+          <!-- <i18n-t v-else keypath="contacts.noResults" /> -->
         </p>
         <div class="mx-4">Page size</div>
 
@@ -241,5 +247,6 @@ watchEffect(async () => {
   if (!currentSegment.value) {
     contactsTotal.value = contactsTable.value.total;
   }
+
 });
 </script>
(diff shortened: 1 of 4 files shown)


rlafuente:
Some good progress, updated the initial comment since it's apparently fixed

rlafuente:
I think the only remaining table to improve is the Callout archive, which is using a simple `<table>` instead of the AppTable component. It's doable but would break the DRY principle.

I can go ahead and try to port this table to AppTable so that it can work with empty sets, but would like assurance that this is the way to go :)

src/modules/contacts/ContactsPage.vue
```diff
@@ -19,6 +19,12 @@
         :items="contactsTable.items"
         class="w-full mt-2 whitespace-nowrap"
       >
+        <template #empty v-if="route.query.s">
+          <p>{{ t('contacts.noResults') }}</p>
```
wpf500: This could be simplified into one slot with the query inline which would be slightly better in terms of not repeating the slot.

```vue
<template #empty>
  <p>{{ route.query.s ? t('contacts.noResults') : t('contacts.noContacts') }}</p>
</template>
```

wpf500 (changes requested):

wpf500:
> I think the only remaining table to improve is the Callout archive, which is using a simple `<table>` instead of the AppTable component. It's doable but would break the DRY principle.
> 
> I can go ahead and try to port this table to AppTable so that it can work with empty sets, but would like assurance that this is the way to go :)

Hmm if I remember correctly the styling of that table is different to other tables? I think its a good idea to use the `AppTable` component but I think this should be an opportunity to unify the styles rather than making the `AppTable` support lots of styles? What do you think? Possibly we could add a `loose` style to support the padding differences between the two?

src/modules/contacts/ContactsPage.vue
rlafuente: ah, elegant! Fixed in 38c4989fe8dbb749e294fa537d675c4992a2631c

rlafuente:
Def, now's always the best time for quick refactor wins :) Just pushed a first working version of the AppTable on the callout page. I still want to review and compare side-by-side with the previous version tomorrow.

rlafuente:
Okay, so the new callout AppTable seems to work just like the previous one, it's ready for review :rocket: 

(in the meantime, I changed the default empty message to a "Loading" message since it can flash for a bit while a page is loading. We should later test the UX more closely on this to be sure that these messages come out as expected)

wpf500:
I think we ultimately need to differentiate between loading and no results, probably by `items` being nullable, but we can push this to another PR as this is an improvement on what came before.

src/modules/contacts/ContactsPage.vue
```diff
@@ -67,7 +70,7 @@
               ><b>{{ n(contactsTable.total) }}</b></template
             >
           </i18n-t>
-          <i18n-t v-else keypath="contacts.noResults" />
+          <!-- <i18n-t v-else keypath="contacts.noResults" /> -->
```
wpf500: I guess this should just be removed now?

wpf500 (approved):

src/modules/contacts/ContactsPage.vue
rlafuente: oof, definitely. Thanks for catching that one (fixed in 8d4eda5)

PR #191, opened 2022-05-18
feat: use join period from URL param, if present

basilesimon:
https://trello.com/c/grDbF1WX/505-add-period-query-param-to-join-page

### DIFF
--- a/src/modules/auth/join/JoinPage.vue
+++ b/src/modules/auth/join/JoinPage.vue
@@ -149,6 +149,12 @@ onBeforeMount(async () => {
     ? Number(route.query.amount)
     : joinContent.value.initialAmount;
 
+  signUpData.period = Object.values(ContributionPeriod).includes(
+    route.query.period
+  )
+    ? (route.query.period as ContributionPeriod)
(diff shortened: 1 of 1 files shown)


src/modules/auth/join/JoinPage.vue
```diff
@@ -149,6 +149,12 @@ onBeforeMount(async () => {
     ? Number(route.query.amount)
     : joinContent.value.initialAmount;
 
+  signUpData.period = Object.values(ContributionPeriod).includes(
+    route.query.period
+  )
+    ? (route.query.period as ContributionPeriod)
+    : ContributionPeriod.Monthly;
```
basilesimon: Setting this instead of a default value as it's what we're doing above in the very definition of `signUpData`.

wpf500: Hmm strange I think that might be a bug as it should be getting set to `joinContent.value.initialPeriod` once `joinContent` has loaded, could you change it?

wpf500 (changes requested):

wpf500 (approved):

PR #192, opened 2022-05-18
Additional info (er, ask for money) for non-members in callout

basilesimon:
[From this card](https://trello.com/c/c7I4SNk4/500-%F0%9F%AA%B3-show-message-if-a-logged-in-user-who-isnt-a-member-tries-to-access-a-member-only-callout): 

> At the moment if a logged in user who isn't a member (e.g. their membership expired) loads a member-only callout they just see the intro text without the form and no prompt.
We should add a prompt which says something like "You need to start a contribution to answer this callout" and point them to the contribution page

image

### DIFF
--- a/src/modules/callouts/pages/CalloutPage.vue
+++ b/src/modules/callouts/pages/CalloutPage.vue
@@ -64,11 +64,7 @@
     <figure class="relative mb-6 pb-[56.25%]">
       <img class="absolute w-full h-full object-cover" :src="callout.image" />
     </figure>
-    <div
-      class="text-lg content-message"
-      v-if="!canSeeButNotRespond"
-      v-html="callout.intro"
-    />
+    <div class="text-lg content-message" v-html="callout.intro" />
     <div
       class="
         w-full
--- a/src/modules/callouts/pages/CalloutPage.vue
+++ b/src/modules/callouts/pages/CalloutPage.vue
@@ -64,7 +64,33 @@
     <figure class="relative mb-6 pb-[56.25%]">
       <img class="absolute w-full h-full object-cover" :src="callout.image" />
     </figure>
-    <div class="text-lg content-message" v-html="callout.intro" />
+    <div
+      class="text-lg content-message"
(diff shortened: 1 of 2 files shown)


src/modules/callouts/pages/CalloutPage.vue
```diff
@@ -64,7 +64,33 @@
     <figure class="relative mb-6 pb-[56.25%]">
       <img class="absolute w-full h-full object-cover" :src="callout.image" />
     </figure>
-    <div class="text-lg content-message" v-html="callout.intro" />
+    <div
```
wpf500: I think it should always show the intro as a teaser. You can see similar behaviour for guests if you look at `<div v-if="showLoginPrompt">` code below

src/modules/callouts/pages/CalloutPage.vue
```diff
@@ -151,6 +177,15 @@ const canRespond = computed(
     currentUser.value?.activeRoles.includes('member')
 );
 
+// edge-case where a member is not contributing
+// (anymore) and should be given some indication as
+// to why they can't reply to a callout
+const canSeeButNotRespond = computed(
+  () =>
+    callout.value?.access === 'member' &&
```
wpf500, line 184: I think this might be missing checking if they are logged in (if not they see a login prompt a la `showLoginPrompt` below. So maybe this should check something like
```ts
callout.value?.access === 'member' && currentUser.value && !currentUser.value.activeRoles.includes('member')
```

wpf500 (changes requested):

wpf500:
Couple of changes but looks good generally, nice one :+1:

src/modules/callouts/pages/CalloutPage.vue
basilesimon: Thank you that's really well spotted

src/modules/callouts/pages/CalloutPage.vue
basilesimon: 👌

PR #198, opened 2022-05-20
WYSIWYG editor

basilesimon:
This PR creates a `<RichTextEditor />` component which you can drop in anywhere and have it return a string of HTML code, i.e. by using `v-model` bindings. It's based on an implementation of [TipTap](https://tiptap.dev/), which I had heard about in visual journalism. Was very much functional out-of-the-box!

image

- The expected keyboard shortcuts work,
- And I've mirrored these basic functionalities with buttons,
    - to which we could add headings, for example

### DIFF
--- a/locales/de.json
+++ b/locales/de.json
@@ -39,6 +39,8 @@
       "endsIn": "Endet in {duration}",
       "startsIn": "startet in {duration}"
     },
+    "toContributionPage": "",
+    "updateContribution": "",
     "youResponded": "Sie haben geantwortet ♥"
   },
   "calloutAdmin": {
@@ -195,6 +197,7 @@
       },
       "tags": "Tags"
     },
+    "noContacts": "",
     "noResults": "Keine Kontakte gefunden",
     "search": "Kontakte durchsuchen...",
     "showingOf": "{start} to {end} von {total} Kontakten werden angezeigt."
@@ -463,6 +466,7 @@
     "passwordInfo": "Das Passwort muss aus mindestens acht Zeichen bestehen und mindestens eine Zahl, einen Groß- und einen Kleinbuchstaben enthalten.",
     "phone": "Telefon",
     "postCode": "Postleitzahl",
+    "richtext": {},
     "saveChanges": "Änderungen speichern",
     "saved": "Gespeichert!",
     "warning": "Ändere hier nur etwas, wenn du das Einverständnis hast. Änderungen werden auch den Nutzern angezeigt."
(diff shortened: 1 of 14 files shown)


wpf500:
Looks good, but I am a bit concerned about the amount of dev effort involved in making our own UI for an editor over using something out-of-the-box. What do you think the benefits are of doing it this way? How long do you think it would take to add the necessary features?

basilesimon:
Get your point.

The feeling I had so far was bringing something that felt very much fresh-out-of-the-box, batteries-included type of thing. And by that I mean that _even I_ only took a few hours to have a reusable editor that we can even customise, that works just really well and looks quite pretty.

Happy to talk (read: list) the necessary features, but so far it's been a breeze.

basilesimon:
- Toolbar is now at the top of the text area
- Toolbar buttons more concise (and er, actual squares)
- Confirmed that pasting in HTML doesn't work. `TipTap` does its own parsing of the elements.

Importantly, creating a link is still WIP.

rlafuente:
This is awesome @basilesimon, and works just great!

Quick QA: 
* after a quick look, I couldn't find the keyboard shortcuts, even though the help text mentions them. I'm used to finding them in a tooltip on the appropriate button, so maybe we could add simple tooltips with name + keyboard shortcuts?
* the "H" button sets a bit of text as a heading, this is notably simpler than h1/h2/h3 selectors. This one would benefit from a tooltip text making it clear what it does.

It's just grand to see the Edit Callout pages with all the bells and whistles, exciting!

wpf500:
@rlafuente I'll add tooltips to describe what buttons do but can we defer adding the shortcuts to a later date?

Shortcuts are dependent on the OS and the WYSIWYG library doesn't provide a nice way to tell which shortcuts are in use. I'm sure it's not that complicated to identify the OS, but seems like hassle for very little gain as the shortcuts are very standard (ctrl+b, etc.).

I'd suggest a simpler approach is to remove the text saying "Select and click the buttons to format the copy, or use keyboard shortcuts." as I don't think it's needed in any case.

rlafuente:
> @rlafuente I'll add tooltips to describe what buttons do but can we defer adding the shortcuts to a later date?

Of course, it's not a key feature at all!

> I'd suggest a simpler approach is to remove the text saying "Select and click the buttons to format the copy, or use keyboard shortcuts." as I don't think it's needed in any case.

Yes definitely -- I was drawn to the shortcut issue because the help text makes that apparent. Removing that reference is a good way to manage expectations.

wpf500:
@rlafuente great, I've pushed the changes to dev. Happy to merge?

wpf500 (approved):

rlafuente:
Yeah :rocket:

PR #204, opened 2022-06-08
Delete callout feature

rlafuente:
The frontend for confirming callout deletion is now usable and good for testing.

Right now there is no operation when confirming deletion -- only a redirect to the callout list.

TODO:
- [x] hook the confirmation action to API endpoint to delete callout
- [x] convert to i18n strings + update locales
- [x] generalise AppModal for more than Confirm/Cancel cases (maybe beyond this PR?)


rlafuente:
I think the only missing detail is to add a success notification, will work on that

rlafuente:
Success notification is up, and as an extra, AppModal now has a `redirectTo` prop in order to avoid the hardcoded URL (see [7774cdb](https://github.com/beabee-communityrm/beabee-frontend/pull/204/commits/7774cdba9cbf39fc89ca23f5cacd29f02417ac6e)). The confirm/cancel buttons are still hardcoded but we can look into that once the need for a more versatile modal comes up.

wpf500:
Just to confirm from our conversation on Slack, I don't think we need `redirectTo` as any redirecting would be handled by the `confirm` handler. Other than looks good to go :+1:

rlafuente:
Done :stars:

PR #207, opened 2022-06-14
feat: replicate callout action

rlafuente:
There's now working callout replication, with the following caveats:

- only works on callouts with a start date (undefined start dates throw validation errors)
- only works on callouts with an image URL (undefined image urls also throw a validation error)
- doesn't properly redirect to the new callout's edit interface (not sure what's tripping the router push)

I think we only need a 5-10min hand-holding call with Will to fix up the remaining bits.

Other TODOs:

- [ ] Add the notification text for a "replicated" event
- [ ] After replicating, perhaps land on the Title & Image page so that the title can be easily edited? (slug needs updating too)


src/modules/callouts/pages/CalloutAdminOverviewPage.vue
```diff
@@ -110,15 +110,29 @@ import {
...
+const router = useRouter();
 const wasJustCreated = route.query.created !== undefined;
 const wasJustUpdated = route.query.updated !== undefined;
 
 const calloutLink = computed(() => `/callouts/${props.callout.slug}`);
+
+async function replicateThisCallout() {
+  var newCalloutData = { ...props.callout };
+  newCalloutData.slug = newCalloutData.slug + '-copy';
+  newCalloutData.title = newCalloutData.title + ' copy';
```
wpf500: This could mutate the `props.callout` property. Although I don't think it actually will be here it's good practice to avoid these patterns if possible, better would be:
```ts
const newCalloutData = {
  ...props.callout,
  slug: props.callout.slug + '-copy',
  title: props.callout.title + ' copy'
  status: undefined // I think this will work as it won't serialise in JSON, but might need to check that
};
```
Note also we never use `var`.

src/modules/callouts/pages/CalloutAdminOverviewPage.vue
```diff
@@ -110,15 +110,29 @@ import {
...
+  delete newCalloutData['status'];
+  const newCallout = await createCallout(newCalloutData);
+  router.push({
+    path: '/admin/callouts/view/' + newCallout.slug,
+    query: { ['replicated']: null },
```
wpf500: `replicated` doesn't need square brackets or quote marks, square brackets are only useful if you want do define a key based on a variable, i.e.
```ts
{replicated: null} === {['replicated']: null}
```

src/modules/callouts/pages/CalloutAdminOverviewPage.vue
```diff
@@ -110,15 +110,29 @@ import {
...
 const calloutLink = computed(() => `/callouts/${props.callout.slug}`);
+
+async function replicateThisCallout() {
+  var newCalloutData = { ...props.callout };
+  newCalloutData.slug = newCalloutData.slug + '-copy';
+  newCalloutData.title = newCalloutData.title + ' copy';
+  delete newCalloutData['status'];
+  const newCallout = await createCallout(newCalloutData);
+  router.push({
+    path: '/admin/callouts/view/' + newCallout.slug,
```
wpf500: Do you think this should go straight to the edit page? (i.e. /admin/callouts/edit/:slug)

wpf500 (changes requested):

wpf500:
Woo replicate callouts :tada:!

I think the replicated callout should always start in draft, so setting both `starts` and `expires` to `null`. Not being able to save `starts` as `null` is a bug, I'll make a PR with a fix.

Callouts should always have an image URL to replicate, the database doesn't allow a null field (although it does allow empty string). Do you have an example of a callout that was failing because of this?

I'm not sure what is tripping the router, we can look together later if that's helpful?

src/modules/callouts/pages/CalloutAdminOverviewPage.vue
rlafuente: Totally -- even better would be to land on the "Title & Image" tab so you can edit the title/slug immediately. How feasible is that?

src/modules/callouts/pages/CalloutAdminOverviewPage.vue
rlafuente: ah! I had added that while bug hunting and it just stayed, thanks for noticing :)

src/modules/callouts/pages/CalloutAdminOverviewPage.vue
rlafuente: Gotcha, still getting to grips with the TS+Vue3 ways of doing things.

rlafuente:
> Callouts should always have an image URL to replicate, the database doesn't allow a null field (although it does allow empty string). Do you have an example of a callout that was failing because of this?

See `/admin/callouts/view/solutions`, which seems to send an empty string in the XHR POST request but causes a validation error: `isUrl: "image must be an URL address"`

I'm testing the router part now and will post another update in a moment

wpf500:
@rlafuente fix is deploying for `starts` validation bug: https://github.com/beabee-communityrm/beabee/pull/149

wpf500:
> Callouts should always have an image URL to replicate, the database doesn't allow a null field (although it does allow empty string). Do you have an example of a callout that was failing because of this?
> 
> See `/admin/callouts/view/solutions`, which seems to send an empty string in the XHR POST request but causes a validation error: `isUrl: "image must be an URL address"`
> 
> I'm testing the router part now and will post another update in a moment

@rlafuente ah I forgot it was validating the URL. This is problem with legacy callouts that weren't created via the API, I need to refresh the data on the dev site as these shouldn't exist any more. You can safely ignore this problem,

src/modules/callouts/pages/CalloutAdminOverviewPage.vue
```diff
@@ -110,15 +111,32 @@ import {
...
+const wasJustReplicated = route.query.replicated !== undefined;
 
 const calloutLink = computed(() => `/callouts/${props.callout.slug}`);
+
+async function replicateThisCallout() {
+  const newCalloutData = {
+    ...props.callout,
+    slug: props.callout.slug + '-copy',
+    title: props.callout.title + ' copy',
+    status: undefined,
```
wpf500, line 132: Changes look good :+1:. Final request is to set the start and expiry date to null to make the new callout a draft:
```ts
{
  ...,
  starts: null,
  expires: null
};
```

wpf500 (changes requested):

src/modules/callouts/pages/CalloutAdminOverviewPage.vue
wpf500: Unfortunately not at the moment as the tabs/steps don't have unique URLs. They definitely should but bit beyond the scope of this PR I think.

src/modules/callouts/pages/CalloutAdminOverviewPage.vue
rlafuente: Done!

rlafuente:
Note about [71ecb2d](https://github.com/beabee-communityrm/beabee-frontend/pull/207/commits/71ecb2d8a57d2cc07d96b861ddcdf11535cae331): 
- there is a warning notification when the user is editing a callout (which ATM doesn't check if it's live, but I'll open a separate issue)
- this notification uses a WarningBanner component that's only meant for this purpose, with a hardcoded message
- since it's a simple wrapper around the AppAlert component, I've edited the code to use AppAlert instead of WarningBanner
- if you agree with this change, we can remove WarningBanner.vue, but I wanted to ensure this is the right move.

wpf500 (approved):

wpf500:
Yeah agreed for removing `WarningBanner`, nice work :+1:

rlafuente:
whoop! :rocket:

PR #220, opened 2022-06-23
Refactor/normalise headers

wpf500:
We have a lot of different header styles creeping in, this is an attempt to rationalise it a bit.

I've merged `SectionTitle` into `AppSubHeading` as they were both being used for the same thing in different places (e.g. your information vs contribution page). I've used `font-semibold` over `font-bold` (the only difference between the two), but happy to go with `font-bold` if you prefer.

We now have:
* `PageTitle`: `font-title text-2.5xl`
* `AppHeading`: `font-title text-xl font-semibold text-body-80`
* `AppSubHeading`: `font-title text-lg font-semibold`

The home page has it's own `SectionTitle` (another one!) which I've cleaned up a bit for consistency but otherwise left it as is for now.

There are probably lots of other places where there are inconsistencies but this is a good start I think

### DIFF
--- a/src/components/AppSubHeading.vue
+++ b/src/components/AppSubHeading.vue
@@ -1,5 +1,5 @@
 <template>
-  <h2 class="font-title font-semibold text-xl text-body-80"><slot /></h2>
+  <h3 class="font-title font-semibold text-lg"><slot /></h3>
 </template>
 <script lang="ts" setup>
 /* Stub script for better type inference */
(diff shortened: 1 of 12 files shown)


rlafuente:
This is great! No blockers on our side.

We'd also like to review the styles to properly match the original specs (there's still some differences), but that task shouldn't block merging this one. Woo :rocket:

PR #225, opened 2022-06-29
feat: show name fields on payment form

wpf500:
Adds first and last name fields to the Stripe payment form. beabee needs this information to be collected separately (Stripe collects it as a full name field) so we can do things like send emails that say "Hi Will"

Note: at the moment there is a slight visual difference between the first/last name fields and the Stripe payment fields because we aren't yet syncing font families into the Stripe iframe. This will be addressed in a future PR

image

### DIFF
--- a/locales/de.json
+++ b/locales/de.json
@@ -553,7 +553,6 @@
     "privacy": "Datenschutzerklärung"
   },
   "joinPayment": {
-    "complete": "Zahlung abschließen",
     "genericError": "Upps! Etwas ist schiefgelaufen. Bitte wenden Sie sich an den Support, sollte das erneut passieren.",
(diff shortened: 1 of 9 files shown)


aiscarvalho:
Looks good!
We noticed the button says "Complete" but it should say something else because there's another step after this one. Tobias had commented about this in our spec :) we changed the button text to "Almost done" in this step and "Complete" in the final step.

wpf500:
@aiscarvalho I'm not sure "Almost done" works for button text, could we use something like "Continue" or "Complete payment" instead?

aiscarvalho:
@wpf500 both sound good; if the payment is finished in this step it's a good idea to use "Complete payment".

PR #227, opened 2022-07-04
Feat/payment history status

wpf500:


rlafuente:
When running locally, after opening the contributions page, I get an empty page with only the "Contribution" heading, along with this repeating error:

image

(I ran `npm install` after it had complained about a missing package, but then got stuck here. If it helps, I'm logged in as a member, not superadmin. **Update:** logging in as superadmin gives the same result)

PR #230, opened 2022-07-07
Membership builder: move account confirmation to its own step

rlafuente:


wpf500:
How do you imagine the background image working with this? Should it show a preview if the background has been changed on the join form or only show saved changes? I think the former makes more sense.

At the moment I'm guessing the account confirmation step isn't showing the background? As the data is currently isolated in each step there is no easy way to show an unsaved background in this step without a hack. However I'm not very happy with how the data is structured at the moment so would be up for changing it to accommodate this.

rlafuente:
Good question -- background wasn't working on the local installation so we overlooked that detail.

I'm not sure that such a detail would warrant a refactor, at least at this moment. Looking at all the conflicting states, maybe we can just remove the background from the confirmation step as a compromise?

wpf500:
Yeah sounds good :+1:

rlafuente:
Okay the Account Confirmation step is now broken out into its own section, but a couple of details emerged:

- I couldn't reliably reproduce this, but often the page loads with an inactive "Update" button, which becomes clickable after any edit on the form
- When there is an error, it only shows after an element is deselected (blur), which is a tiny bit confusing since the "Update" button is made inactive when e.g. clearing a required field, but no feedback. 

Not sure how to quickly solve all of these, but at the very least we could include a notice next to "Update" saying there are errors in the form. Ideally with a tooltip which lists all the problems, maybe?

rlafuente:
Another update! This is turning out to be more of a general review, but the result will be :star2: _snappy_ :star2: New stuff: 
- Now there is an additional status message next to the Update button for when there are form errors. It would benefit from a copywriting review since it mentions "fields above" on the string list, but that's another concern
- Fields now show error state on update (not just unfocus) which gives more explicit and immediate feedback

https://user-images.githubusercontent.com/159868/178709422-2119452d-07eb-4b33-9b3d-e7aefee002a2.mp4

Open points:

- Regarding my previous remark _"often the page loads with an inactive "Update" button, which becomes clickable after any edit on the form"_, i think this had to do with a blank background image, which is not a valid state. A simple way out would be to  have a default image loaded, because UX-wise it's a bit confusing to not be able to proceed.
- the "The value is required" string comes directly from the validator, but I'm not sure how to hook this with translations (and could use clearer copy, e.g. "This field cannot be empty")
- the alignment for the error messages in the "Account confirmation" step is off (too much whitespace above), just making a note not to forget to review this

rlafuente:
OK I think this is ready for review+merge!

src/modules/membership-builder/components/steps/AccountConfirmation.vue
```diff
@@ -0,0 +1,187 @@
...
+watch(validation, () => {
+  emit('update:error', validation.value.$errors.length > 0);
+  emit('update:validated', !validation.value.$invalid);
+});
+
+async function handleUpdate() {
+  if (setupContent.value) {
+    await Promise.all([
+      updateContent('join/setup', setupContent.value),
+      updateContent('general', {
```
wpf500: I think this probably shouldn't be here any more

src/modules/membership-builder/MembershipBuilderPage.vue
```diff
@@ -5,7 +5,11 @@
         <font-awesome-icon :icon="['fa', 'check-circle']" />
         {{ t('form.updated') }}
       </span>
-      <span v-if="dirty" class="mr-4">
+      <span v-else-if="validation.$invalid" class="mr-4 text-danger">
+        <font-awesome-icon :icon="['fa', 'warning']" />
+        {{ t('form.errors.aggregator') }}
```
wpf500: Bit confusing that it says "check the fields above", we should add another string which is "check for errors below" or something

wpf500 (commented):
Great nice one, a couple of changes which I can just sort out then I'll merge :+1:

wpf500:
Replying to your points @rlafuente 

> Open points:
> 
>  * Regarding my previous remark _"often the page loads with an inactive "Update" button, which becomes clickable after any edit on the form"_, i think this had to do with a blank background image, which is not a valid state. A simple way out would be to  have a default image loaded, because UX-wise it's a bit confusing to not be able to proceed.

I think this is actually because the button is disabled when nothing on the form has changed. This is confusing so I'll remove that feature.

> 
> * the "The value is required" string comes directly from the validator, but I'm not sure how to hook this with translations (and could use clearer copy, e.g. "This field cannot be empty")
> 
> * the alignment for the error messages in the "Account confirmation" step is off (too much whitespace above), just making a note not to forget to review this

Both of these are on my radar as something I want to address in an upcoming PR to make `AppInput` etc. much easier to use.
* The validation is overly complicated and boilerplatey at the moment. In the beabee-signup [AppInput](https://github.com/beabee-communityrm/beabee-signup/blob/main/src/components/AppInput.vue) all the validation is declared within the component so you just add a `required` attribute and it does the rest, leading to [much simpler code](https://github.com/beabee-communityrm/beabee-signup/blob/main/src/pages/index.vue). With a few changes we could have all the validation and localisation logic in one place rather than in every form, big win!
* The margin error is because the `mb-4` class is applied to the input (there is no wrapping div)

For now I'll fix the margin error with a wrapping div.

rlafuente:
+1 on every point, thanks for the thorough explanations!

PR #231, opened 2022-07-07
Feat/collapse nav

aiscarvalho:
image

A collapsed version of the main menu, visible between breakpoints md and xl.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2507,14 +2507,20 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001282",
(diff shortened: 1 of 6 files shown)


wpf500:
Looks great :+1:. Do you think we need a tooltip or something to describe the icons? Or it could expand on hover? Although neither of these solutions work for tablets. I'll merge this and put something on the backlog about making the icons clearer

aiscarvalho:
@wpf500 we agree that some sort of text label would be important; also agree with reviewing the icons for extra clarity.

PR #233, opened 2022-07-13
fix: removed callout image height constrain

aiscarvalho:
In the callout page, the main image had a fixed ratio which resulted in an undesirable crop of the original image.
This is a small fix: the image takes the full width but has variable height.

A callout with a _portrait_ image:
image

A callout with a _landscape_ image:
image


wpf500:
Looks good :+1:

Only minor thing is this is failing the prettier check (code formatting). You should be able to set up your IDE so it auto runs prettier on save, otherwise you can run `npx prettier -w .` in the project root

PR #245, opened 2022-08-22
Design review

rlafuente:
This PR is for a set of visual corrections and updates based on the spec and general perception of things that could work better. 

We'll update this with further comments once we get back from holiday, along with proper screenshots and a list of changes.

### Forms (before/after)

beforeafter

### Join form

image


rlafuente:
I can't pin down the reason for the Prettier failure -- running locally gives no errors

src/components/pages/admin/contacts/ContactTag.vue
```diff
@@ -11,16 +11,16 @@
       px-1
       mr-1
       mb-1
-      last-of-type:mb-0
       text-xs
       border border-white
+      select-none
     "
-    :class="hasRemoveIcon ? 'hover:border-primary-40' : ''"
+    :class="hasRemoveIcon ? '' : ''"
```
wpf500: This could just be removed I think?

src/components/tabs/AppVTabsItem.vue
```diff
@@ -6,10 +6,11 @@
       justify-between
       items-baseline
       p-1
-      mb-1
+      pl-2
+      pr-2
+      pb-2
```
wpf500: Probably better expressed as `p-2 pt-1` (why is top padding different?)

src/components/tabs/AppTabs.vue
```diff
@@ -22,7 +22,8 @@
             flex
             items-center
             justify-center
-            hover:font-bold
```
wpf500, line 25: I'm very happy to see the bold hover going :tada:.

I had to add in extra elements here to make the bold hover work without layout problems so that can now also be removed. Everything from line 12 to 32 can now probably just be something like (not tested):

```vue
<span class="text-body-80 hover:text-body">
  {{ item.label }}
  <span v-if="item.count !== undefined" class="ml-1">
    ({{ item.count }})
  </span>
</span>
```

wpf500 (changes requested):

wpf500:
@aiscarvalho @rlafuente I have concerns about using 40% opacity colours for text.

If you enable Firefox's Accessibility tools then you can see that it fails the background/foreground colour contrast test and while that's definitely not the only accessibility problems we have I'd rather not introduce new ones.

With theming we can't completely eliminate contrast problems (although we should probably advise users about it at some point), but I don't think 40% opacity text will ever be accessible so I think we need a different option here (and probably aiming to remove all `text-x-40` classes going forwards

rlafuente:
> @aiscarvalho @rlafuente I have concerns about using 40% opacity colours for text.

You're totally right, 40% was just there for disabled buttons. We're thinking of adding a 60% tone for cases where lighter text would be desirable (which right now is titles and help text)

src/components/tabs/AppVTabsItem.vue
rlafuente: Ah good, we're still wrapping our heads around Tailwind's way to do things. Padding is slightly different on top to better frame the text, since using the same padding on top and bottom results (in our ways) in an apparent imbalance between the top and bottom whitespace.

src/components/pages/admin/contacts/ContactTag.vue
rlafuente: Good find! Yes

src/components/tabs/AppVTabsItem.vue
rlafuente: And while trying it out, `p-2` does work best, so let's just use that

src/components/AppHeading.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <h2 class="font-title font-semibold text-xl text-body-80"><slot /></h2>
+  <h2 class="font-title font-semibold text-xl text-body-40"><slot /></h2>
```
wpf500: @rlafuente `AppHeading` are using 40% opacity text too

wpf500: Ah I see from your comment you meant you want to introduce a 60% opacity for these components?

rlafuente: yes that's it!

rlafuente:
note to also fix #263 before merging

src/components/tabs/AppTabs.vue
aiscarvalho: Fixed here: https://github.com/beabee-communityrm/beabee-frontend/pull/245/commits/49c3c6dedce172a28ad040ba81e8f4b3f0eb4800#diff-d9f58ec032801d1531ea86aa8c89542c7c332fb516447d51b08b70bdb852fdeb

wpf500 (approved):

PR #253, opened 2022-08-25
Fix Stripe appearance

wpf500:


wpf500:
This still doesn't fix the font family unfortunately, it's surprisingly difficult to figure out how to tell Stripe where to get the fonts from. I'm going to merge this as it's better than before and will look into the fonts later

PR #269, opened 2022-09-27
Callout improvements

wpf500:
### Added admin preview for callouts

Admins need to be able to see their callouts even when they aren't open. At the moment if they load a callout page they can't see the form.

This PR turns the Preview/View button into an admin "view mode" and adds a Participate button to open the callout in "participate mode", i.e. the same mode as if they went via the callouts list (`/callouts`)

image

When in view mode the submit button is disabled and there is a small notification to the admin:

image

### Always show thanks message if there's been a response

At the moment we only show the thanks message when they submit a response, if you then reload the page it disappears:

image

I think it is a nicer experience and more consistent to always show the thank you message:

image


aiscarvalho:
These are great additions that make a lot of sense; some comments on these:

- View ended callouts: could we show the callout form but have it disabled instead? We can keep the button as "View" and use the disabled form state to avoid submissions after the end date.

- We think admins should always be allowed to View/Preview callouts, even if it's members only. Maybe we disable the submission button on view/preview modes to avoid admins that aren't members from responding?
And if so, should we consider adding a "Reply" button to present admins/members?
image
We don't want to overcomplicate things but we think it makes sense to separate adminy tasks like creating/testing/previewing and the act of replying to a callout.

- Thank you message: agreed!

wpf500:
### Preview for ended callouts

The main use case for having a "preview" mode for ended callouts (rather than just a disabled form) is for callouts with conditional logic, i.e. for the form can change depending on a particular answer. The other day at the Cable I wanted to see how a callout we'd sent out previously worked, but because it was disabled I couldn't interact with the options. In the end I just duplicated the callout and tested it in a fresh callout.

What do you think about this? Maybe I'm overcomplicating things and creating a duplicate is fine if you want to test stuff out?

### View/preview callouts

If I have understood you correctly then I think I totally agree

At the moment my flow is this:
image

So the button changes depending on if the callout is open or not, this is confusing!

I think you are suggesting this:
image

So clicking View/Preview in the admin view is consistent (always goes to "Preview" mode), and there is also a button that takes them to the "Normal" mode.

Can you confirm I've understood?

aiscarvalho:
Yes, that was the suggestion! The scheme it much simpler to explain the flow :brain: 

Regarding the preview for ended callouts, I don't think you're overcomplicating. It's always nice to see how an ended callout worked even if it's old/ended. Ans would be great to allow users to do that without having to create a duplicate.

How about if we just disable the submission button? The form is all clickable but there's no way to submit, and we can include a note. Something like this:
image

wpf500:
@aiscarvalho this is now ready for review I think.

I've chosen to use "Participate" rather than "Reply" for the button as this is consistent with the language we've used elsewhere.

Note because the submit button is controlled by the Form.IO plugin I can't easily make it look disabled (with lower opacity) or put the hint text next to it. It is however disabled and I figure it's okay with it just below?

aiscarvalho:
@wpf500 I agree that participate is a better label for the button. Great updates!

aiscarvalho (approved):
We've just went through the changes and everything looks good. All of these are great improvements to the management and experience of callouts.

PR #280, opened 2022-10-03
Use callout summary card on admin view

wpf500:
It's currently quite hard to easily see the status of a callout on the admin view page. Is the following callout open or closed? You need to read the dates and see if they are before/after now.

image

I suggest reusing the component we have on the admin dashboard. It shows the dates and status in an easily accessible way, and by reusing a component I think it is more consistent and familiar.

image


rlafuente:
this is a great improvement, no comments!

rlafuente (approved):

PR #282, opened 2022-10-05
Advanced search

wpf500:
Implementation of advanced search!

image

Changes from the design spec:
* Hiding save segment for now (is in the next piece of work)
* Added paginated result display (count and pages) above table
* Removed Cancel button in favour of Reset which resets the query to either the current segment or all results
* Moved Search button to the left as without Save Segment it looked strange by itself and we have a UI pattern of "<Action> | <Cancel>" (in this case Search | Reset)

As part of this work there have been a few improvements to components:
* `AppSelect` now supports option groups
* `AppInput` now supports a prefix (e.g. €)
* `AppRadioGroup` has better alignment, supports validation and can auto generate `name` attribute values
* Introduced `AppPaginatedResult` for a consistent UI for `Paginated` tables


aiscarvalho:
Great work with the advanced search!

While testing we found that querying by "Active role is not" appears to not be working (we get the same results with "is" and "is not"):

_Active role is Superadmin_ vs _Active role is not Superadmin_ returns the same results
Screenshot 2022-11-14 at 15-49-21 Community - beabee dev
Screenshot 2022-11-14 at 15-49-04 Community - beabee dev

_Active role is Member_ vs _Active role is not Member_ returns the same results
Screenshot 2022-11-14 at 15-51-21 Community - beabee dev
Screenshot 2022-11-14 at 15-50-46 Community - beabee dev

And we have two of suggestions to improve the UX:
- Ignore empty querys/rules. Currently the search button is inactive if there's an empty rule and it's a behavior that isn't clear at first:
image

- When building a rule for **Member since** the default is _Member since is dd-mm-yyyy_ which translates to _Member since **exact date is** dd-mm-yyyy_.   
Would it be possible to replace _is/is not_ by _exact date is/exact date is not_?  
Alternatively we'd suggest switching the default operator to _after/before_ since it's a bit more intuitive. Same thing for _Contribution cancelled exact date is dd-mm-yyyy_ and _Member until exact date is dd-mm-yyyy_
image

wpf500:
@aiscarvalho I've fixed the "Active role" filter, thanks for pointing it out. All of the filters "Active role", "Active member", "Member since" and "Member expires" will all be replaced soon with a role filter that is much more intuitive.

I've changed "is" to "is on" rather than "exact date is" as I think it conveys the same meaning and slightly cleaner, what do you think?

I can remove the validation on selecting a filter, but it will still cause the button to be disabled when they choose a filter until they fill in the arguments:
image

I did try this before but Tobi and I found it a bit strange, what should it do when this happens? Still ignore them?
image

aiscarvalho:
> @aiscarvalho I've fixed the "Active role" filter, thanks for pointing it out. All of the filters "Active role", "Active member", "Member since" and "Member expires" will all be replaced soon with a role filter that is much more intuitive.

:+1: 
 
> I've changed "is" to "is on" rather than "exact date is" as I think it conveys the same meaning and slightly cleaner, what do you think?

Perfect!
 
> I can remove the validation on selecting a filter, but it will still cause the button to be disabled when they choose a filter until they fill in the arguments:

Don't remove validation! I think I mis-explained.
In this example, where the filter is almost complete, it's nice to show the input validation. It's a quick way to hint why the "search" button is disabled.

> I did try this before but Tobi and I found it a bit strange, what should it do when this happens? Still ignore them? image

Agreed. Then just disregard our suggestion, let's keep it as is :)

PR #283, opened 2022-10-05
Add, edit and delete roles

rlafuente:
Here's a first working version with a good set of caveats to be solved before merging:

- We're openly violating DRY by having a New Role form and an Edit Role form that are pretty similar (though with key differences). The parallel "visible" variables are also clunky, though everything seems to work
- Not sure if the `editRole` reactive var is the way to go here, but again, works
- When editing a role, the form does not reflect its current state: it doesn't show its start-end dates, and we're still trying to figure out proper binding with a custom component (`AppDateInput`, made to deal with Date<->String roundtrips), but editing the form and saving does update the role with the new values. 
- When creating a new role we're forced to set `dateExpires`, even though we have a "never expires" option. We just monkeypatched it by setting it 10 yrs after `dateAdded`, but surely there's a sane way to solve this.
- After a change, the roles list is not updated and needs refreshing for seeing the new state. I'm sure this is obvious to fix but not sure how.
- The "Delete" button just deletes without confirmation (for the sake of quickly implementing, then tweaking). We want to review the UX here
- The "Add Role" button takes up too much space and attention, we'll find another place for it

Tomorrow we'll review the design details and update this.


src/components/forms/AppDateInput.vue
```diff
@@ -0,0 +1,33 @@
+<template>
```
wpf500: I'm not sure about this component. I tried something like this for the callout editor and found that it was much more reliable just to keep the date in text YYYY-MM-DD then convert before sending to the API.

Arguably there might be a place for an `AppDateTimeInput` which could handle converting the `Date` to a date/time string and vice versa but for now I think you should follow the pattern used in the callout edit code

src/pages/admin/contacts/[id]/index.vue
```diff
@@ -121,18 +121,188 @@ meta:
...
+            >Add role</AppButton
+          >
+        </div>
+      </div>
+
+      <div
+        v-if="isNewRoleFormVisible"
+        class="flex flex-1 justify-center rounded border border-primary-20 py-4 px-10"
+      >
+        <form @submit.prevent="handleNewRoleFormSubmit" class="flex-initial">
```
Comment (wpf500) [inline src/pages/admin/contacts/[id]/index.vue]: I think all of these forms should be broken out into separate components so they can handle their own state rather than the page needing lots of confusing variables (`isNewRoleFormVisible`, `isEditRoleFormVisible`, etc.).

Ideally the notes editor would be split out too as now there is more than 1 form on the page it would be better for them all to be separate.

src/utils/api/member.ts
```diff
@@ -98,6 +105,17 @@ export async function updateMember(
   return deserializeMember(data);
 }
 
+export async function addMemberRole(
```
wpf500: This should be removed I think?

wpf500 (commented):
I've left a few comments but not sure what to review in detail at this stage as not sure what you've still got planned. As you've already said the edit and add form should definitely be the same form (where in "add" mode you can select a role).

wpf500:
> When creating a new role we're forced to set dateExpires, even though we have a "never expires" option. We just monkeypatched it by setting it 10 yrs after dateAdded, but surely there's a sane way to solve this.

I think this is probably because you need to pass it `null` if it doesn't expire?

The type definition for updating a role is:
```ts
export interface UpdateMemberRoleData {
  dateAdded: Date;
  dateExpires: Date | null;
}
```

rlafuente:
I think we got to a proper component architecture:
- RoleEditorItem for each displayed role entry, containing the role edit form and delete function
- RoleEditor containing all items, plus the new role form

However I'm still struggling with some final details where your guidance is needed:
- Forms are submitting successfully but the frontend isn't updated with the new values. I'm not sure if I'm misunderstanding how reactive properties are supposed to work, but I've tried updating the proxy values on save and there's no visible change. Because of this, any changes are only visible on reload.
- Form fields don't start with set values (e.g. when editing, have the form fields pre-filled to the existing values). I tried setting the `value` property but I'm not sure why it doesn't have the expected behaviour, maybe there's some details I overlooked in our form components?

Other details before this is definitely done:
- Delete acts immediately without confirmation. We'd like to come up with a custom component for cases like this where confirmation is needed, but for now a modal might suffice.
- Add a "no end date" option to the edit form (or leave it as it is?)
- Review translatable strings and remove hardcoded labels

src/pages/admin/contacts/[id]/index.vue
```diff
@@ -205,37 +201,34 @@ const contactAnnotations = reactive({
   tags: [] as string[],
 });
 const securityLink = ref('');
+const newRoleFormVisible = ref(false);
```
Comment (wpf500) [inline src/pages/admin/contacts/[id]/index.vue]: Can be removed :tada:

src/pages/admin/contacts/[id]/index.vue
```diff
@@ -196,7 +192,7 @@ const contact = ref<GetMemberDataWith<
   'profile' | 'contribution' | 'roles'
 > | null>(null);
 const contactTags = ref<string[]>([]);
-const loading = ref(false);
+const noteFormLoading = ref(false);
```
Comment (wpf500) [inline src/pages/admin/contacts/[id]/index.vue]: Would be tempted to revert this to keep the PR changes simple

src/pages/admin/contacts/[id]/index.vue
```diff
@@ -187,6 +182,7 @@ import RichTextEditor from '../../../../components/rte/RichTextEditor.vue';
 formatLocale;
 
 const { t, n } = useI18n();
+const inputT = (key: string) => t('contacts.data.rolesCopy.' + key);
```
Comment (wpf500) [inline src/pages/admin/contacts/[id]/index.vue]: I think this isn't used any more?

src/components/pages/admin/contacts/RoleEditorItem.vue
```diff
@@ -0,0 +1,158 @@
...
+const editRole = reactive({
+  role: '',
+  startDate: '',
+  startTime: '',
+  endDate: '',
+  endTime: '',
+});
+
+const { t, n } = useI18n();
+const inputT = (key: string) => t('contacts.data.rolesCopy.' + key);
```
wpf500: Generally prefer using the full `t(..)` unless it's really being used a lot as keeping full i18n keypaths means the keys are more greppable

src/components/pages/admin/contacts/RoleEditorItem.vue
```diff
@@ -0,0 +1,158 @@
...
+}
+
+function parseDateTime(date: string, time: string): Date {
+  return new Date(date + 'T' + time);
+}
+
+onBeforeMount(async () => {
+  loading.value = false;
+  editRole.role = props.role.role;
+  editRole.startDate = props.role.dateAdded.toLocaleDateString();
```
wpf500: I'm fairly sure this will need to be `format(YYYY-MM-DD)` and `format(HH:mm)` for date and time respectively as I think the date/time inputs expect ISO8601 style formats

src/components/pages/admin/contacts/RoleEditorItem.vue
```diff
@@ -0,0 +1,158 @@
...
+          >Edit</AppButton
+        >
+        <AppButton variant="dangerText" size="sm" @click="handleDeleteRole">
+          Delete
+        </AppButton>
+      </div>
+    </div>
+
+    <div v-if="formVisible" class="flex flex-1 p-4">
+      <form @submit.prevent="handleFormSubmit" class="flex-initial">
```
wpf500: I guess the form could probably be extracted to it's own component so it can be shared between add/edit

wpf500 (commented):
Definitely getting there! Nice that it doesn't really touch the `admin/contacts/[id]/index.vue` page any more which is the aim :tada:

wpf500:
> I think we got to a proper component architecture:
> 
>     * RoleEditorItem for each displayed role entry, containing the role edit form and delete function
> 
>     * RoleEditor containing all items, plus the new role form
> 
> 
> However I'm still struggling with some final details where your guidance is needed:
> 
>     * Forms are submitting successfully but the frontend isn't updated with the new values. I'm not sure if I'm misunderstanding how reactive properties are supposed to work, but I've tried updating the proxy values on save and there's no visible change. Because of this, any changes are only visible on reload.

I think I would solve this by emitting an event to the parent page telling it something has changed and getting it to refetch the data. The only problem is that it might look a bit janky while the API response loads. The alternative is to emit `edit` and `add` events with the new role data so the parent can update it's list of roles.

> 
>     * Form fields don't start with set values (e.g. when editing, have the form fields pre-filled to the existing values). I tried setting the `value` property but I'm not sure why it doesn't have the expected behaviour, maybe there's some details I overlooked in our form components?

This might just be because the inputs are in the wrong format? See my comment above

> 
> 
> Other details before this is definitely done:
> 
>     * Delete acts immediately without confirmation. We'd like to come up with a custom component for cases like this where confirmation is needed, but for now a modal might suffice.
> 
>     * Add a "no end date" option to the edit form (or leave it as it is?)

Needs a no end date option
> 
>     * Review translatable strings and remove hardcoded labels

rlafuente:
Updates are properly shown now! I've also fixed the time fields. How does it look on your side?

The only remaining point for a minimally acceptable UI is to add a delete confirmation when you click the delete button (right now it's instant).

It would be desirable to add some smoothness to the transitions with simple animations, but maybe that's a task for the backlog.

rlafuente:
The last chores (add confirmation dialog to delete, review translation) are done! 

The only detail remaining is to componentize the add/edit form, which is still a bit of work. We're placing this feature in QA and maybe address this in the next sprint, or is it a blocker?

src/components/pages/admin/contacts/RoleEditor.vue
```diff
@@ -0,0 +1,167 @@
...
+import AppRadioGroup from '../../../../components/forms/AppRadioGroup.vue';
+import RoleEditorItem from '../../../../components/pages/admin/contacts/RoleEditorItem.vue';
+import { GetMemberData } from '../../../../utils/api/api.interface';
+import { updateRole } from '../../../../utils/api/member';
+import { onBeforeMount, ref, reactive } from 'vue';
+import { useI18n } from 'vue-i18n';
+
+const emit = defineEmits(['update']);
+const props = defineProps<{
+  contact: GetMemberData;
```
wpf500: You probably want `GetMemberDataWith<'roles'>` instead of `GetMemberData` here, will fix the `vue-tsc` error above

src/components/pages/admin/contacts/RoleEditor.vue
```diff
@@ -0,0 +1,167 @@
...
+const newRole = reactive({
+  role: '',
+  startDate: '',
+  startTime: '',
+  endDate: '',
+  endTime: '',
+});
+
+// FIXME: hardcoded
+const roleOptions = [
```
wpf500, line 133: These need to be internationalised properly, you can use `t('common.role.<id>')` for that

src/components/pages/admin/contacts/RoleEditorItem.vue
```diff
@@ -0,0 +1,187 @@
...
+}
+
+function parseDateTime(date: string, time: string): Date {
+  return new Date(date + 'T' + time);
+}
+
+onBeforeMount(async () => {
+  loading.value = false;
+  editRole.role = props.role.role;
+  editRole.startDate = props.role.dateAdded.toISOString().split('T')[0];
```
wpf500: Would prefer using date formatting rather than string splitting, see [callout code](https://github.com/beabee-communityrm/beabee-frontend/blob/main/src/pages/admin/callouts/new.vue#L133)
```ts
editRole.startDate = format(props.role.dateAdded, 'yyyy-MM-dd');
editRole.startTime = format(props.role.dateAdded, 'HH:mm');
```

src/components/pages/admin/contacts/RoleEditor.vue
```diff
@@ -0,0 +1,167 @@
...
+
+const { t } = useI18n();
+const inputT = (key: string) => t('contacts.data.roles.' + key);
+
+const loading = ref(false);
+const formVisible = ref(false);
+const roleHasStartDate = ref(false);
+const roleHasEndDate = ref(false);
+const newRole = reactive({
+  role: '',
```
wpf500, line 126: This needs to be of type `PermissionType`. Probably make a new interface e.g.
```ts

interface EditableRole {
  role: PermissionType | '';
  startDate: string;
  startTime: string;
  endDate: string;
  endTime: string
}

...

const newRole = reactive<EditableRole>({
  role: '',
  startDate: '',
  startTime: '',
  endDate: '',
  endTime: ''
}
```

Then in `handleFormSubmit` you would just check that role is set to something other than empty string, e.g.
```ts
if (!newRole.role) return;
```

src/components/pages/admin/contacts/RoleEditor.vue
```diff
@@ -0,0 +1,167 @@
...
+              </div>
+              <div>
+                <AppInput v-model="newRole.endTime" type="time" required />
+              </div>
+            </div>
+          </div>
+        </div>
+
+        <div class="my-3 flex py-3">
+          <AppButton type="submit" variant="primary" :loading="loading">{{
```
wpf500: The button should not be enabled unless the form validates. In practice this is very easy to achieve as long as all the validation props (`required`, etc.) are set correctly which they appear to be here.

```vue
<AppButton ... :disabled="validation.$invalid" />
```

In the script
```ts
const validation = useVuelidate();
```

wpf500:
> The last chores (add confirmation dialog to delete, review translation) are done!
> 
> The only detail remaining is to componentize the add/edit form, which is still a bit of work. We're placing this feature in QA and maybe address this in the next sprint, or is it a blocker?

This can be done as a separate card in the sprint if you like, but I think it does need to be addressed

rlafuente:
Almost there! But vue-tsc complains about this:

```
src/components/pages/admin/contacts/RoleEditor.vue:149:40 - error TS2345: Argument of type 'string' is not assignable to parameter of type 'PermissionType'.

149     await updateRole(props.contact.id, newRole.role, {
                                           ~~~~~~~~~~~~

src/components/pages/admin/contacts/RoleEditorItem.vue:136:40 - error TS2345: Argument of type 'string' is not assignable to parameter of type 'PermissionType'.

136     await updateRole(props.contact.id, editRole.role, {
                                           ~~~~~~~~~~~~~
```

I'm not sure how to deal with the mismatching types here, since `newRole.role` is supposed to be a string. What am I missing?

src/components/pages/admin/contacts/RoleEditor.vue
```diff
@@ -0,0 +1,177 @@
...
+import AppInput from '../../../../components/forms/AppInput.vue';
+import AppButton from '../../../../components/forms/AppButton.vue';
+import AppSelect from '../../../../components/forms/AppSelect.vue';
+import AppRadioGroup from '../../../../components/forms/AppRadioGroup.vue';
+import RoleEditorItem from '../../../../components/pages/admin/contacts/RoleEditorItem.vue';
+import {
+  GetMemberDataWith,
+  MemberRoleData,
+} from '../../../../utils/api/api.interface';
+import { EditableRole } from 'roles.interface.ts';
```
wpf500: See `vue-tsc` error, this import shouldn't have the extension (don't add .js either)

src/components/pages/admin/contacts/RoleEditor.vue
```diff
@@ -0,0 +1,177 @@
...
+<script lang="ts" setup>
+import AppLabel from '../../../../components/forms/AppLabel.vue';
+import AppInput from '../../../../components/forms/AppInput.vue';
+import AppButton from '../../../../components/forms/AppButton.vue';
+import AppSelect from '../../../../components/forms/AppSelect.vue';
+import AppRadioGroup from '../../../../components/forms/AppRadioGroup.vue';
+import RoleEditorItem from '../../../../components/pages/admin/contacts/RoleEditorItem.vue';
+import {
+  GetMemberDataWith,
+  MemberRoleData,
```
wpf500: Unused import needs removing

src/components/pages/admin/contacts/roles.interface.ts
```diff
@@ -0,0 +1,7 @@
+interface EditableRole {
```
wpf500: If you want it in a separate file (a good idea) you need to export it.
```ts
export interface EditableRole {
  ...
}
```

src/components/pages/admin/contacts/roles.interface.ts
```diff
@@ -0,0 +1,7 @@
...
+  role: PermissionType | '';
```
wpf500, line 4: Need to import this type, VSCode or similar should give you a quick fix to import it from the right place

wpf500 (commented):
Close, just a few finicky type problems!

wpf500:
I think your type problems should be resolved if you fix the `EditableRole` definition

PR #288, opened 2022-10-10
feat: add not found page

wpf500:
Note the email address is set to whatever that newsroom's support email is

image


aiscarvalho (approved):
All good!

PR #306, opened 2022-11-14
Contact contribution edit form

rlafuente:
There's a working form ready to test now. There is some validation left to implement, but I'm not sure of the exact use cases under which the form should not allow the user to proceed. 

Looking at it, the form also looks pretty bare, some help texts could clarify the one-word fields like Source or Reference.

You mentioned a way to prepopulate the Source field from the API, can you remind me how to go about this?


rlafuente:
I just saw the vue-tsc complaints, will take care of that early Wednesday

src/components/pages/profile/contribution/EditManualContribution.vue
```diff
@@ -0,0 +1,149 @@
+<template>
```
wpf500: This component should be under `src/components/pages/admin/contacts` as it's only for the `admin/contacts`. The idea of `src/components/pages` is to put route specific components under a path that is similar to the route they are used in

src/components/pages/profile/contribution/EditManualContribution.vue
```diff
@@ -0,0 +1,149 @@
...
+  <form @submit.prevent="handleUpdate">
+    <AppHeading class="mt-6 mb-2">
+      {{ t('contribution.billing') }}
+    </AppHeading>
+
+    <AppSelect
+      v-model="contribution.type"
+      label="Contribution Type"
```
wpf500: Needs localising, quite a few of these to do

src/components/pages/profile/contribution/EditManualContribution.vue
```diff
@@ -0,0 +1,149 @@
...
+const isActiveMember = computed(
+  () => contribution.membershipStatus === MembershipStatus.Active
+);
+const isManualActiveMember = computed(
+  () => isActiveMember.value && contribution.type === ContributionType.Manual
+);
+
+// Only non-active members and monthly manual contributors can change their period
+// as otherwise proration gets complicated
+const showChangePeriod = computed(
```
wpf500: There's no need to hide this for manual contributors so `isActiveMember`, `isManualActiveMember` and `showChangePeriod` can all go

src/components/pages/profile/contribution/EditManualContribution.vue
```diff
@@ -0,0 +1,149 @@
...
+    !isActiveMember.value ||
+    (isManualActiveMember.value &&
+      contribution.period !== ContributionPeriod.Annually)
+);
+
+watch(
+  toRef(props, 'id'),
+  async (id) => {
+    const member = await fetchMember(id, ['contribution']);
+    contribution.type = member.contribution.type;
```
wpf500: You need to check if the type is `ContributionType.None` or `ContributionType.Manual`, otherwise you risk showing this form for automatic contributors.

src/components/pages/profile/contribution/EditManualContribution.vue
```diff
@@ -0,0 +1,149 @@
...
+    contribution.amount = member.contribution.amount;
+    contribution.period = member.contribution.period;
+    contribution.source = member.contribution.paymentSource.source;
```
wpf500: `paymentSource` can be different things, so the compiler needs to be sure it's a `ManualPaymentSource` before giving you access to `source` and `reference`

The definition of the type is:
```ts
export type PaymentSource =
  | StripeCardPaymentSource
  | GoCardlessDirectDebitPaymentSource
  | StripeBACSPaymentSource
  | StripeSEPAPaymentSource
  | ManualPaymentSource;
```
`PaymentSource` is a [discriminated union](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions), which means it has a property which tells you which of the possible subtypes it actually is. 

In this case that property is called `method`, you can see this because all the types listed have a unique value for `method` (see [this bit in api.interface.ts](https://github.com/beabee-communityrm/beabee-frontend/blob/feat/contact-contrib/src/utils/api/api.interface.ts#L100-L139)). So to be sure that `paymentMethod` is a `ManualPaymentSource` you need to check if `method` is `null`:
```ts
const paymentSource = member.contribution.paymentSource;
if (paymentSource?.method === null) {
  contribution.source = paymentSource.source || '';
  contribution.reference = paymentSource.reference || '';
} else {
  // In practice this shouldn't be possible as you've checked in the type is Manual or None above
  contribution.source = '';
  contribution.reference = '';
}

File: src/components/pages/profile/contribution/EditManualContribution.vue (hunk 4 above)
Comment (wpf500) [reply src/components/pages/profile/contribution/EditManualContribution.vue]: I would be tempted to move this whole bit outside of `EditManualContribution` and only load the component if they are a Manual or None type contributor

**Review by wpf500 — CHANGES_REQUESTED** [review]

**Comment by wpf500** [comment]
@rlafuente we can pair programme on some of these bits if you want, especially around the payment source etc.

For reference to validation and functionality the form needs you should refer to the legacy app version, e.g.
https://dev.beabee.io/members/e04e090a-c803-4e48-aa1e-25210b68fdfc/contribution

**Comment by wpf500** [comment]
You can get the list of available payment sources from:
```ts
const manualPaymentSources = (await fetchContent('contacts')).manualPaymentSources;
```

rlafuente:
> we can pair programme on some of these bits if you want, especially around the payment source etc.

Yeah, this would definitely help. Let me tackle the points you mentioned above and I'll ping you about the rest!

rlafuente:
Okay, I've done my best to tackle the tsc errors, but there are 3 remaining points that I need guidance with:

```
src/components/pages/admin/contacts/EditManualContribution.vue:15:7 - error TS2322: Type 'number | null' is not assignable to type 'string | number | undefined'.
  Type 'null' is not assignable to type 'string | number | undefined'.

15       v-model="contribution.amount"
         ~~~~~~~~~~


src/components/pages/admin/contacts/EditManualContribution.vue:33:7 - error TS2322: Type 'string | undefined' is not assignable to type 'string | null'.
  Type 'undefined' is not assignable to type 'string | null'.

33       v-model="contribution.source"
         ~~~~~~~~~~


src/components/pages/admin/contacts/EditManualContribution.vue:116:5 - error TS2322: Type 'ContributionType' is not assignable to type 'ContributionType.Manual | ContributionType.None'.

116     contribution.type = member.contribution.type;
        ~~~~~~~~~~~~~~~~~


Found 3 errors in the same file, starting at: src/components/pages/admin/contacts/EditManualContribution.vue:15
```

src/pages/admin/contacts/[id]/contribution.vue
```diff
@@ -0,0 +1,39 @@
...
+    </Suspense>
+  </div>
+</template>
+<script lang="ts" setup>
+import { onBeforeMount, ref } from 'vue';
+import { GetMemberDataWith } from '../../../../utils/api/api.interface';
+import { ContributionType } from '@beabee/beabee-common';
+import EditManualContribution from '../../../../components/pages/admin/contacts/EditManualContribution.vue';
+
+const showEditForm = ref(false);
```
Comment (wpf500) [inline src/pages/admin/contacts/[id]/contribution.vue]: This is what `computed` is great at

```ts
const showEditForm = computed(() => 
  props.contact.contribution.type === ContributionType.Manual ||
  props.contact.contribution.type === ContributionType.None
)
```

src/pages/admin/contacts/[id]/contribution.vue
```diff
@@ -0,0 +1,39 @@
...
+  pageTitle: menu.community
+  role: admin
+</route>
+
+<template>
+  <div class="grid lg:grid-cols-2 xl:grid-cols-3">
+    <Suspense>
+      <EditManualContribution v-if="showEditForm" :id="contact.id" />
+      <p v-else>
+        Editing a member's contribution is only possible when their contribution
```
Comment (wpf500) [inline src/pages/admin/contacts/[id]/contribution.vue]: Needs localising

wpf500 (changes requested):

wpf500:
@rlafuente I pushed a few small changes before consulting with you, sorry about that. They were:
1. Merging with main as it had merge conflicts
2. Fixing a broken reference to a i18n string
3. Fixing broken margins between elements by moving the margins from `AppInput` to a parent element, currently `AppInput` doesn't support margins directly (the margins are placed on the wrong element)
4. Adding a currency prefix to the amount (as per the legacy app reference form)

The form needs two further changes to reflect the behaviour in the legacy app, they are:
1. The contribution/payment fields should disappear if the admin selects contribution type None
  a. The frontend should then send empty values for amount and period to the API, e.g.
```ts
forceUpdateContribution({type: ContributionType.None, amount: undefined, period: undefined})
```
2. The admin should be able to select a blank payment source

https://user-images.githubusercontent.com/2084823/204784146-0559b394-684a-46ff-8dff-469347b05c01.mp4

saschafoerster:
Just a sidenote: I see for all of them periods. I also want to add single time donations, if that is possible. Right now I have to choose yearly or monthly and I can't add a date.

rlafuente:
Thank you for going ahead with all the fixes! I'll look into this when we pick it up next Wednesday.

wpf500 (approved):

PR #309, opened 2022-11-16
Save/update segments

wpf500:
Saving segments, either creating a new one or updating an existing one

image

image


aiscarvalho (approved):
Great work!

After testing I was thinking that we're just missing a way to delete segments.
The quickest fix would be to show a "delete" button, together with "save" button - save is visible when you edit filters, delete would also be visible upon editing (mocked below):

image

Should I add this to the backlog?

wpf500:
@aiscarvalho deleting segments is already in the backlog under the Segments feature: https://trello.com/c/jrSX7jCl/441-%F0%9F%8E%81-segments

PR #313, opened 2022-11-16
Merge a couple of callout steps and make draft editing work better

wpf500:
* Merge the "Title and image" step with the "URL and sharing" step. Everything in the latter was dependent on the former, and when editing a callout (rather than creating) the step had very little in it.
* Make the permalink editable when a callout is in draft or scheduled status

Before:
image
image

Now for new callouts:
image

And updating callouts:
image


aiscarvalho:
Everything looks good! Makes a lot more sense now, that we have both steps together :)

PR #321, opened 2022-11-30
Pagination review

rlafuente:
We wanted to review the design of the pagination component.

Before:

image

After:

image

And a more compact version for the payments history:

image

We still want to simplify a bit of the code, but would appreciate feedback on the UX here!


wpf500:
Much nicer! I won't comment on the code yet but I think this is a great improvement UX wise :)

rlafuente:
The vue-tsc checks are failing due to a couple of type inconsistencies, and I can't make much of it... tried some fixes but can't get to a proper result.

rlafuente:
Other than the vue-tsc checks, the feature seems to work properly and is otherwise ready for review+merge

rlafuente:
All fixed and ready for QA!

src/components/AppPaginatedResult.vue
```diff
@@ -45,6 +70,11 @@ const currentPageSize = computed({
   set: (newPageSize) => emit('update:pageSize', newPageSize),
 });
 
+const pageSizes = [12, 25, 50, 100].map((x) => ({
+  id: x,
+  label: x + ' per page',
```
wpf500: Needs internationalising. Note that translation is "reactive", meaning it should be wrapped in a `computed` so it updates if the locale changes, e.g. something like (not tested)

```ts
const pageSizes = computed(() => [12, 25, 50, 100].map(x => ({
  id: x,
 label: t('the.key.here', {page: n(x)})
}))
```

Note that the page number is passed as a `page` property and is i18ned itself using the `n` i18n function which you can import with
```ts
const { t, n } = useI18n();
```
This doesn't make any difference for small numbers but 1000 would become "1,000" (in EN), and it's always good practice to i18n numbers in case there is a language out there than does something with "100"

wpf500 (changes requested):
There's also an unused variable warning, can you remove `pages` if it's not being used?

src/components/AppPagination.vue
```diff
@@ -1,45 +1,59 @@
 <template>
-  <ul class="-mx-2 flex items-center">
-    <li>
-      <button
-        class="p-2"
-        :class="isFirst && 'opacity-25 hover:border-link hover:font-bold'"
+  <ul class="flex justify-start">
+    <li class="-mr-px">
```
wpf500, line 3: Ah nice, I didn't know `-mr-px` was a thing. Time to fix some of my other code!

wpf500:
@rlafuente looks good :+1:

wpf500 (approved):

PR #322, opened 2022-11-30
Add contact page

wpf500:
This PR adds the contact add page.

It includes refactoring the `RoleEditor` and `EditManualContribution` to be more generic so they can be used on both the contact page and the add contact page

image


aiscarvalho:
We tested the new feature and it's working well.

What we missed the most was a clear hint of how to edit the new contact once it's created. Since some contact elements are editable in the overview page (role and about) it's not clear that name and email are editable in the information tab.
As a quick fix we would suggest to add an "Edit" link, like we had in the spec. We think there's a backlog card already to address the editing incoherence of the single contact page.  
Spec screenshot:
image

Other comments that crossed our mind while testing but aren't blockers (except the bug :):
- Possible bug: after creating  a new contact we see the overview, information and contribution tabs but, if navigate to the contact list and then return to the newly created contact page we only see the overview and information tabs. The information tab doesn't open on click (it worked when we right clicked and picked "open in a new tab").
- Would it make sense to include all contact fields when creating a new contact: delivery opt-in, address, description, notes, tag?
- Would it make sense to allow the admin to pick a "join date"?
- In the overview page, it would be nice to hide Preferred channel, Phone and Groups if the contact doesn't have them since they're not editable fields. Newsletter could also be hidden or include a note saying it's not editable in beabee but synced from MC (we figure). At the moment, added contacts show _Newsletter: none_ which is not ideal:

image

aiscarvalho:
Something I forgot to mention above, I tried creating a contact with an existing email and it shows an empty error box:
image

wpf500:
@aiscarvalho I've fixed the lack of error when adding a contact with a duplicate email address :+1: . I haven't been able to reproduce the tabs problem I think the most likely explanation is that dev was mid deploy.

I'll capture the changes to the contact overview page on a separate card as it doesn't directly relate to this PR.

> - Would it make sense to include all contact fields when creating a new contact: delivery opt-in, address, description, notes, tag?
> - Would it make sense to allow the admin to pick a "join date"?

@TobiasHauswurz do you have any thoughts here? I think it's better to keep the form simple. Again any suggestions I will put in a card in the backlog.

wpf500:
I'm merging this now as all other comments will be handled in future cards :)

PR #325, opened 2022-12-07
feat: split general settings into multiple forms

wpf500:
Split the general settings form into multiple forms, this is helpful for users who don't want to fill everything in yet and improves useability with less scrolling to find the update button

image


rlafuente:
This is MUCH smoother, wonderful! All good here.

(the one thing that stands out -- and I'll add a backlog issue for this -- is that the General tab doesn't yet make use of the two-column layout, and a single column for many fields is somewhat wasteful. We'll look into this later)

PR #331, opened 2022-12-14
Create notice form

wpf500:


JonasSchatz:
@wpf500 can't add you as a reviewer, since you created the PR in the first place. But... you know what to do. Roast me :)

src/pages/admin/notices/view/[id].vue
```diff
@@ -0,0 +1,82 @@
...
+const router = useRouter();
+
+const showDeleteModal = ref(false);
+
+onBeforeMount(async () => {
+  notice.value = await fetchNotice(props.id as string);
+});
+
+const confirmDeleteNotice = () => {
+  deleteNotice(props.id);
```
Comment (wpf500) [inline src/pages/admin/notices/view/[id].vue]: This should have an `await` to ensure the delete has been successful before redirecting

src/components/callout/CalloutSummary.vue
```diff
@@ -56,10 +52,13 @@ import {
   GetCalloutData,
   GetCalloutDataWith,
 } from '../../utils/api/api.interface';
+
+import AppButton from '../forms/AppButton.vue';
```
wpf500: A couple of unused imports

src/pages/admin/notices/edit/[id].vue
```diff
@@ -0,0 +1,150 @@
...
+    ? format(notice.value.starts, 'HH:mm')
+    : '';
+  data.expirationDate = notice.value.expires
+    ? format(notice.value.expires, 'yyyy-MM-dd')
+    : '';
+  data.expirationTime = notice.value.expires
+    ? format(notice.value.expires, 'HH:mm')
+    : '';
+  data.text = notice.value.text;
+  data.buttonText = notice.value.buttonText ? notice.value.buttonText : '';
```
Comment (wpf500) [inline src/pages/admin/notices/edit/[id].vue]: What do you think of this vs using the `||` operator?
```ts
data.buttonText = notice.value.buttonText ? notice.value.buttonText : '';
data.buttonText = notice.value.buttonText || '';
```

I like the simplicity of the latter although understand that for people coming from other languages it can be confusing. It is however very common in JS/TS

src/pages/admin/notices/edit/[id].vue
```diff
@@ -0,0 +1,150 @@
...
+    starts: new Date(data.startDate + 'T' + data.startTime),
+    expires: new Date(data.expirationDate + 'T' + data.expirationTime),
+    text: data.text,
+    buttonText: data.buttonText,
+    url: data.url,
+  };
+  if (notice.value) {
+    return await updateNotice(notice.value.id, noticeToSubmit);
+  } else {
+    return await createNotice(noticeToSubmit);
```
Comment (wpf500) [inline src/pages/admin/notices/edit/[id].vue]: Not sure I understand this as this page can only ever be loaded for editing a notice.

`saveNotice` can't be called unless notice is defined as the form won't show, although unfortunately the compiler doesn't know this. I would usually solve this with a quick return at the top of the method and a comment

```ts
async function saveNotice() {
  if (!notice.value) return; // Can't submit without notice being loaded
  ...
}
```

src/pages/callouts/[id].vue
```diff
@@ -132,7 +132,7 @@ import InfoMessage from '../../components/InfoMessage.vue';
 
 import 'formiojs/dist/formio.form.css';
 import { useRoute } from 'vue-router';
-import CalloutStatus from '../../components/callout/CalloutStatus.vue';
+import ItemStatusComponent from '../../components/item/ItemStatus.vue';
```
Comment (wpf500) [inline src/pages/callouts/[id].vue]: The convention is to give component imports the same name as their file, so this should be
```ts
import ItemStatus from '../../components/item/ItemStatus.vue';
```

src/utils/api/notice.ts
```diff
@@ -1,7 +1,24 @@
...
-import { GetNoticeData, GetNoticesQuery, Serial } from './api.interface';
+import {
+  CreateNoticeData,
+  GetNoticeData,
+  GetNoticesQuery,
+  Serial,
+} from './api.interface';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function deserializeNotice(notice: any): any {
```
wpf500: I don't think this needs to use type `any`, it could be
```ts
function deserializeNotice(notice: Serial<GetNoticeData>): GetNoticeData {
  ...
}
```

The idea is that the compiler will throw an error if you forget to deserialize a property, e.g.
```ts
function deserializeNotice(notice: Serial<GetNoticeData>): GetNoticeData {
  return notice;
  // Type 'Serial<GetNoticeData>' is not assignable to type 'GetNoticeData'.
  //   Types of property 'createdAt' are incompatible.
  //     Type 'string' is not assignable to type 'Date'.ts(2322)
}
```

This is especially useful in the future as we could for example add a new `Date` property to `GetNoticeData`, then the compiler would automatically catch that we need to deserialize the date.

At the moment there are a few `any` deserializers (e.g. `deserializeContact`), but that's just because I haven't figured out how to write proper generic types for them to work.

wpf500 (commented):
Generally looks great :tada:. Just a few small comments and change requests.

I would like to see the edit and add page using a shared form component, rather than replicating the form on `edit.vue` and `new.vue`, but am happy to defer that to a future task and doesn't need to be addressed in this PR.

src/pages/admin/notices/view/[id].vue
Comment (JonasSchatz) [reply src/pages/admin/notices/view/[id].vue]: Done

src/pages/admin/notices/edit/[id].vue
Comment (JonasSchatz) [reply src/pages/admin/notices/edit/[id].vue]: I'll use it, I was looking for a pattern like this :)

src/pages/callouts/[id].vue
Comment (JonasSchatz) [reply src/pages/callouts/[id].vue]: Problem was: There is already an `ItemStatus` in `common`. Renamed the component to something a bit more descriptive

Comment (wpf500) [reply src/pages/callouts/[id].vue]: Ahh sorry I missed this :+1:

JonasSchatz:
I've implemented all your change requests, also refactored the form in its own component. Needed to read up on how to pass values back from children to parents.

JonasSchatz:
@wpf500 I've implemented all your change requests, also refactored the form in its own component. Needed to read up on how to pass values back from children to parents.

src/components/notice/NoticeForm.vue
```diff
@@ -0,0 +1,79 @@
...
+
+<script lang="ts" setup>
+import { reactive } from 'vue';
+import { useI18n } from 'vue-i18n';
+import AppForm from '../forms/AppForm.vue';
+import AppInput from '../forms/AppInput.vue';
+import AppLabel from '../forms/AppLabel.vue';
+import { NoticeFormData } from './notice.interface';
+
+const emit = defineEmits(['submit']);
```
wpf500: `emit` is never used, you don't need to store the value returned from `defineEmits` in this case

src/pages/admin/callouts/view/[id]/index.vue
```diff
@@ -132,13 +132,13 @@ const wasJustUpdated = route.query.updated !== undefined;
 
 const showDeleteModal = ref(false);
 
-const confirmDeleteCallout = () => {
+async function confirmDeleteCallout() {
   deleteCallout(props.callout.slug);
```
Comment (wpf500) [inline src/pages/admin/callouts/view/[id]/index.vue]: Needs an await on `deleteCallout`

wpf500 (commented):
Passing the data back with a submit event is definitely the way to go here :+1:

I think I'd be tempted to make `NoticeForm` take a `notice` parameter with type `GetNoticeData | undefined`, then convert that into `data` inside the `NoticeForm`. This way you could remove the empty `data` definition for new notice page and all the boilerplate on the edit page, and have it just inside `NoticeForm`.

You could also convert the data back into `CreateNoticeData` before emitting `submit`, that makes the component interface nicely symmetrical (get in a `GetNoticeData`, get out a `CreateNoticeData`) and means you don't need to expose the internal type `NoticeData`.

wpf500:
Lovely lovely :tada:

@JonasSchatz  I have added [a commit](https://github.com/beabee-communityrm/beabee-frontend/pull/331/commits/f95911fdf71ecbb2220fa03dcb7bf0bdb5537e4a) which adds the breadcrumb component to your pages, I thought it would be simpler to show how to do it with a diff. You'll probably notice there is a fair amount of duplication at the moment but I'm planning to address that later.

PR #336, opened 2023-01-04
chore: remove edit button in favour of link to view

wpf500:
Small UI papercut change we agreed before Christmas: admins probably don't want to go directly to editing a callout from the dashboard, but they might want to view it. This PR removes the button in favour of being able to click on the box and goes to `/view/` instead of `/edit/`

Before:

image

After:

image


aiscarvalho (approved):
Everything looks good!

PR #339, opened 2023-01-04
chore: rename information to account

wpf500:
Another UI papercut, renaming everything that says "Information" to "Account"


JonasSchatz (approved):
Approved :)

PR #341, opened 2023-01-05
chore: remove commitlint and unused stylelint

wpf500:
We discussed whether to keep commitlint and we decided to remove it.

Processes agreed:
* Remove commitlint so we can go back to commit messages that don't have to fit into arbitrary categories
* Merge PRs in instead of squashing so we maintain commit history

I'd like to suggest one more rule which is that pull request titles do use the commitlint style format. I think the categories actually make sense for PR titles as they are automatically pulled into the release notes and are actually quite helpful for understanding what the PR does (is it a fix/new feature/chore/etc.)


rlafuente:
+1! 

Reason to be cheerful :^)

JonasSchatz (changes requested):
Looking good, I couldn't find any more places where husky and/or commitlint are used in the repo now. 
Suggestion: To emphasize using the commitlint-style PR-titles, we might want to add a GitHub Pull Request template (https://axolo.co/blog/p/part-3-github-pull-request-template) as a reminder for the existing options.

wpf500:
Good idea :+1:. I've added a template, what do you think @JonasSchatz ?

JonasSchatz:
> Good idea 👍. I've added a template, what do you think @JonasSchatz ?

Yes, this can be merged :)

JonasSchatz (approved):

PR #344, opened 2023-01-05
Chore/settings general 2 col layout

aiscarvalho:
The new layout in different screen widths
 
image
image
image

Other layout improvements:
- styled the "delete" button in other links the same way as the delete button in search filters
image
image
- adjusted the tab component paddings
- adjusted spacing for the horizontal rulers separating the 3 different forms
- added border and adjusted spacing in the image upload component
image

### DIFF
--- a/src/components/forms/AppImageUpload.vue
+++ b/src/components/forms/AppImageUpload.vue
@@ -3,7 +3,7 @@
     <AppLabel v-if="label" :label="label" :required="required" />
     <div class="flex items-start gap-4">
       <div
-        class="flex-0 basis-28 rounded border border-primary-40 bg-primary-20"
+        class="flex-0 basis-28 overflow-hidden rounded border border-primary-40 bg-primary-20"
       >
         <img
           :src="imageUrl"
--- a/src/components/forms/AppImageUpload.vue
+++ b/src/components/forms/AppImageUpload.vue
@@ -2,7 +2,9 @@
   <div>
     <AppLabel v-if="label" :label="label" :required="required" />
     <div class="flex items-start gap-4">
-      <div class="flex-0 basis-28 bg-grey-light">
+      <div
+        class="flex-0 basis-28 rounded border border-primary-40 bg-primary-20"
+      >
         <img
           :src="imageUrl"
           :width="width"
--- a/src/components/tabs/AppTabs.vue
+++ b/src/components/tabs/AppTabs.vue
@@ -1,9 +1,5 @@
 <template>
-  <AppSelect
-    v-model="selectedItem"
-    :items="items"
-    class="mb-4 md:mb-6 md:hidden"
-  />
+  <AppSelect v-model="selectedItem" :items="items" class="mb-4 md:hidden" />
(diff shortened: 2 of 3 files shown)


src/components/tabs/AppTabs.vue
```diff
@@ -1,10 +1,16 @@
 <template>
-  <AppSelect v-model="selectedItem" :items="items" class="mb-4 md:hidden" />
-  <ul class="mb-4 hidden border-b border-primary-40 text-sm md:-mx-2 md:flex">
+  <AppSelect
+    v-model="selectedItem"
+    :items="items"
+    class="mb-4 md:mb-6 md:hidden"
```
wpf500: I don't think you need `md:mb-6` as `md:hidden` will hide the element in any case,

src/pages/admin/settings/index.vue
```diff
@@ -40,18 +40,18 @@ meta:
         </div>
       </AppForm>
 
-      <div class="my-6 border-b border-b-primary-40" />
+      <div class="my-8 block border-b border-b-primary-40" />
```
wpf500: I think the `<div>` will be a `block` in any case so not sure this is needed?

src/pages/admin/settings/index.vue
```diff
@@ -84,18 +84,18 @@ meta:
           />
         </div>
       </AppForm>
-
-      <div class="my-6 border-b border-b-primary-40" />
-
+    </div>
+    <div>
+      <div class="my-8 block border-b border-b-primary-40 md:hidden" />
```
wpf500: Same as above, probably no need for `block`

src/components/forms/AppImageUpload.vue
```diff
@@ -2,7 +2,9 @@
   <div>
     <AppLabel v-if="label" :label="label" :required="required" />
     <div class="flex items-start gap-4">
-      <div class="flex-0 basis-28 bg-grey-light">
+      <div
+        class="flex-0 basis-28 rounded border border-primary-40 bg-primary-20"
```
wpf500: It's really small but if you look closely you can see that the image is overlapping the rounded border. You want to add either `overflow-hidden` here or `rounded` to the `` so it respects the rounding, the former is probably preferable in this instance

wpf500 (changes requested):

src/components/tabs/AppTabs.vue
aiscarvalho: Good catch; I missed that bit but it's now gone.

src/pages/admin/settings/index.vue
aiscarvalho: You're correct; deleted `block`.

src/components/forms/AppImageUpload.vue
aiscarvalho: Yes! Added `overflow-hidden`, thanks! I saw the tiny overlapping edge and forgot to fix it.

wpf500 (approved):

PR #392, opened 2023-04-25
fix: disable track initial page view

wpf500:
The initial page view always seems to be the root (`/`) regardless of the page the user landed on. The real initial page is triggered subsequently anyway so just remove the extra initial page view


wpf500:
Meant to raise PR for fork

PR #407, opened 2023-05-30
feat: add "end now" button to callout view

rlafuente:
In the callout view, the new "End now" button sets the expiry date to the current date/time. The user is returned to the callout list with a success message, indicating that the change can be reverted by editing the callout.

~~(i have vue-formio errors on `npm run check` but I bet this has to do with my installation and is unrelated to this feature.)~~ not anymore

### Screenshots

image

## Checklist before merging

- [ ] Translations for all new i18n strings


rlafuente:
A couple of notes after testing this feature:

1. Thinking that it might make sense to have a "Reopen" button for closed callouts, useful to fix a misplaced click, or to quickly make a closed callout available again. It could set the callout as open without an expiry date, and the success notification ought to mention this clearly ("if you want to set an expiry date, edit the callout")
2. Not sure about also including the "End now" button in the callout edit page; the "Dates" section is one click away once you open the edit UI. Adding an extra action button (besides Update and Revert to Draft) will likely make the interface clunkier. 
3. After closing a callout with the "End now" button in the callout detail page, we're redirected to the callout list. Would it make more sense to remain on the detail page?

wpf500:
1. Sounds good
2. Agreed
3. I think it should stay on the same page but that is slightly complicated by the need to update the UI to set the status to ended. This is a pattern that generally needs looking at (e.g. same for contact updates) but there isn't currently a pattern to reuse for this. So either: 1) just leave it as a redirect for now or 2) let's talk about the implementation.

rlafuente:
We've just added the "Reopen" action to closed callouts. 

Right now it changes the start date to the moment when it was reopened, not sure if it's the ideal action.

(also, agree that we can leave the post-action redirect until it becomes urgent to revise the UI patterns)

src/pages/admin/callouts/view/[id]/index.vue
```diff
@@ -134,6 +150,37 @@ async function confirmDeleteCallout() {
   router.push({ path: '/admin/callouts' });
 }
 
+async function endThisCallout() {
+  const now = new Date();
+  const calloutData = {
```
Comment (wpf500) [inline src/pages/admin/callouts/view/[id]/index.vue]: I think you can probably simplify this to:
```ts
await updateCallout(props.callout.slug, {expires: new Date()})
```

i.e. You only need to send an update for the properties that are changing

src/pages/admin/callouts/view/[id]/index.vue
```diff
@@ -134,6 +150,37 @@ async function confirmDeleteCallout() {
...
+  addNotification({
+    title: t('calloutAdmin.ended'),
+    variant: 'success',
+  });
+  router.push({ path: '/admin/callouts' });
+}
+
+async function reopenThisCallout() {
+  const now = new Date();
+  const calloutData = {
```
Comment (wpf500) [inline src/pages/admin/callouts/view/[id]/index.vue]: Hmm not sure whether to update the `starts` property or not. I'd be tempted to go for the less "destructive" option here and not overwrite it as they can always update that in the edit interface if they want to.

Also same as above, can probably be simplified to
```ts
await updateCallout(props.callout.slug, {expires: null})
```

wpf500 (changes requested):

src/pages/admin/callouts/view/[id]/index.vue
Comment (rlafuente) [reply src/pages/admin/callouts/view/[id]/index.vue]: aha great -- since `calloutData` was set using `props.callout`'s properties in `replicateThisCallout` (which was the function i repurposed), i thought they had to be included. Fixing

Comment (rlafuente) [reply src/pages/admin/callouts/view/[id]/index.vue]: hm, i get

```Argument of type '{ expires: Date; }' is not assignable to parameter of type 'AllowNull<CalloutData & CalloutFormData>'.
  Type '{ expires: Date; }' is missing the following properties from type 'AllowNull<CalloutData & CalloutFormData>': slug, title, excerpt (....)
```

and if i add `...props.callout` to have these properties, i get a 400 error with `property status should not exist`, which seems to be because `status` comes with props.callout (which i had sorted before by setting it as undefined in calloutData before feeding it to `updateCallout`)

i guess i can try removing it, but in the end it seems messier than what we had with `calloutData`. I'm sure there's a simple way out but I can't grok it easily

src/pages/admin/callouts/view/[id]/index.vue
Comment (rlafuente) [reply src/pages/admin/callouts/view/[id]/index.vue]: same problem as above :/

src/pages/admin/callouts/view/[id]/index.vue
Comment (wpf500) [reply src/pages/admin/callouts/view/[id]/index.vue]: Hmm the type needs updating on the `updateCallout` method. The general pattern we use is to create a new type called `UpdateCalloutData` in `api.interface.ts` then use that type for the update method:

```ts

// In api/api.interface.ts

// Find this line
export type CreateCalloutData = AllowNull<CalloutData & CalloutFormData>;
// Add this new line
export type UpdateCalloutData = Partial<CreateCalloutData>;

// In api/callout.ts

// Replace this
export async function updateCallout(
  slug: string,
  calloutData: CreateCalloutData
): Promise<GetCalloutData> {

// With

export async function updateCallout(
  slug: string,
  calloutData: UpdateCalloutData
): Promise<GetCalloutData> {
```

wpf500 (approved):

PR #415, opened 2023-06-14
refactor: move callout admin components to pages/admin/callouts

wpf500:
Move the admin specific callout components into the correct folder, move `ActionButton` into general components.


wpf500:
Changes applied in #437

PR #459, opened 2023-09-06
feat: add conditional link to sidebar logo

rlafuente:
This one is straightforward; only question is whether the hardcoded URLs are okay to use?

### DIFF
--- a/src/components/AppLogo.vue
+++ b/src/components/AppLogo.vue
@@ -1,10 +1,10 @@
 <template>
-  <a :href="canAdmin ? '/admin' : '/profile'">
+  <router-link to="/">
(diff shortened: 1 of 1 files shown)


wpf500:
You should use a `<router-link>` component to use the internal Vue router, and link to `/` as that *should* redirect to the right place based on if the user is an admin or not.

```vue
<router-link to="/">
...
</router-link>
```

rlafuente:
oh right -- fixed!

wpf500:
Thanks. Can you also remove the unused import `canAdmin`?

rlafuente:
d'oh, fixed

PR #493, opened 2023-10-11
feat: callout map intro

rlafuente:
We still want to tune the styling before it's ready to merge, but there's a couple of practical questions that are worth mentioning right away:

- it seems overkill to use a state variable (`introOpen`) but I couldn't get the widget to close otherwise. I'm sure there's a more elegant approach here
- there seems to be an inconsistency with callout properties: the `intro` property seems to be identical to `excerpt` in the map view, but it's a different text in the callout page. We wanted to feature the intro text but it's not present in the available callout properties. We can showcase this issue tomorrow if needed
- we couldn't test the gallery view because the callout reported `Error: Callout does not have a gallery`

### DIFF
--- a/src/components/pages/callouts/CalloutIntroPanel.vue
+++ b/src/components/pages/callouts/CalloutIntroPanel.vue
@@ -1,9 +1,12 @@
 <template>
   <CalloutSidePanel :show="true" @close="$emit('close')">
     <div class="lg:flex lg:flex-col lg:justify-center lg:items-start h-full">
-      <AppTitle class="pt-8 mb-4">{{ $props.callout.title }}</AppTitle>
-      <div class="mb-6 text-lg" v-html="$props.callout.intro"></div>
-      <AppButton @click="$emit('close')" variant="primary" class="px-6">
+      <AppTitle class="pt-8 mb-4">{{ callout.title }}</AppTitle>
+      <figure class="mb-6">
+        <img class="w-full object-cover" :src="callout.image" />
+      </figure>
+      <div class="mb-6 text-lg" v-html="callout.intro"></div>
+      <AppButton variant="primary" class="px-6" @click="$emit('close')">
         {{ t('callout.getStarted') }}
       </AppButton>
     </div>
--- a/src/components/pages/callouts/CalloutIntroPanel.vue
+++ b/src/components/pages/callouts/CalloutIntroPanel.vue
@@ -1,17 +1,19 @@
 <template>
   <CalloutSidePanel :show="true" @close="$emit('close')">
-    <AppHeading class="mb-4">{{ $props.callout.title }}</AppHeading>
-    <div v-html="$props.callout.intro"></div>
-    <AppButton @click="$emit('close')">
-      {{ t('callout.getStarted') }}
-    </AppButton>
+    <div class="lg:flex lg:flex-col lg:justify-center lg:items-start h-full">
+      <AppTitle class="pt-8 mb-4">{{ $props.callout.title }}</AppTitle>
+      <div class="mb-6 text-lg" v-html="$props.callout.intro"></div>
+      <AppButton @click="$emit('close')" variant="primary" class="px-6">
+        {{ t('callout.getStarted') }}
+      </AppButton>
+    </div>
   </CalloutSidePanel>
 </template>
 
 <script lang="ts" setup>
 import { GetCalloutDataWith } from '../../../utils/api/api.interface';
 import CalloutSidePanel from './CalloutSidePanel.vue';
-import AppHeading from '../../AppHeading.vue';
+import AppTitle from '../../AppTitle.vue';
 import AppButton from '../../button/AppButton.vue';
 import { useI18n } from 'vue-i18n';
 
--- a/src/components/pages/callouts/CalloutIntroPanel.vue
+++ b/src/components/pages/callouts/CalloutIntroPanel.vue
@@ -1,7 +1,7 @@
 <template>
   <CalloutSidePanel :show="true" @close="$emit('close')">
     <AppHeading class="mb-4">{{ $props.callout.title }}</AppHeading>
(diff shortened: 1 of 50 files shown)


src/components/pages/callouts/CalloutIntroPanel.vue
```diff
@@ -0,0 +1,22 @@
+<template>
+  <CalloutSidePanel :show="true" @close="$emit('close')">
+    <AppHeading class="mb-4">{{ $props.callout.title }}</AppHeading>
```
wpf500: You don't need `$props`, props are made globally available in the `<template>`

```vue
<AppHeading class="mb-4">{{ callout.title }}</AppHeading>
```

Could you remove all the `$props` references for consistency with the wider codebase?

src/pages/callouts/[id]/gallery.vue
```diff
@@ -54,6 +54,11 @@ meta:
       @close="router.push({ hash: '' })"
       @click.stop
     />
+    <CalloutIntroPanel
+      :callout="callout"
+      :show="!!introOpen"
+      @close="handleCloseIntro"
```
Comment (wpf500) [inline src/pages/callouts/[id]/gallery.vue]: I would just inline `introOpen = false` here, it's probably clearer than having a method for such a simple thing

src/pages/callouts/[id]/gallery.vue
```diff
@@ -54,6 +54,11 @@ meta:
       @close="router.push({ hash: '' })"
       @click.stop
     />
+    <CalloutIntroPanel
+      :callout="callout"
+      :show="!!introOpen"
```
Comment (wpf500) [inline src/pages/callouts/[id]/gallery.vue]: You shouldn't need the `!!` I think?

src/components/pages/callouts/CalloutIntroPanel.vue
```diff
@@ -0,0 +1,22 @@
+<template>
+  <CalloutSidePanel :show="true" @close="$emit('close')">
```
wpf500: Should this not be being bound to a prop called `show`? Does the panel ever close at the mo?

src/pages/callouts/[id]/map.vue
```diff
@@ -409,6 +421,8 @@ onBeforeMount(async () => {
   responses.value = (await fetchResponsesForMap(props.id)).items.filter(
     (r): r is GetCalloutResponseMapDataWithAddress => !!r.address
   );
+
+  introOpen.value = true;
```
Comment (wpf500) [inline src/pages/callouts/[id]/map.vue]: Could you not just initialise `introOpen` with `true` then you don't need to do this?

wpf500 (changes requested):
Looks great in general, nice one :+1:.

Could you add the callout image to the intro panel too, I guess in between the title and intro as per the standard callout page

src/components/pages/callouts/CalloutIntroPanel.vue
wpf500: Or maybe this is working because Vue is automatically passing the `show` prop from the parent? In which case I think it would be less confusing to pass the `show` value explicitly.

wpf500:
Not sure what you mean about the intro text so maybe we can discuss tomorrow :)

rlafuente:
Thanks! We'll fix these today

> Not sure what you mean about the intro text so maybe we can discuss tomorrow :)

I can't reproduce the behaviour I (thought I) saw yesterday so just ignore that for now

src/components/pages/callouts/CalloutIntroPanel.vue
rlafuente: Makes total sense. I added the prop `show` to CalloutIntroPanel, and changed the element property to `:show="show"` -- just making sure this is the right approach?

src/pages/callouts/[id]/gallery.vue
```diff
@@ -91,6 +99,10 @@ const selectedResponse = computed(() => {
   }
 });
 
+function handleCloseIntro() {
```
Comment (wpf500) [inline src/pages/callouts/[id]/gallery.vue]: I think this one still needs to be inlined

wpf500 (changes requested):
Thanks for addressing the changes, one more tiny one!

What happens if you click on a response with the intro panel open? Does it go over the top? I have a feeling that you need to set `introOpen` to false if the user tries to open a response?

rlafuente:
Ok good one -- so I've set the intro to close when a response is viewed, but also when the Add Response button is clicked

src/components/pages/callouts/CalloutIntroPanel.vue
```diff
@@ -0,0 +1,27 @@
+<template>
+  <CalloutSidePanel :show="show" @close="$emit('close')">
+    <div class="lg:flex lg:flex-col lg:justify-center lg:items-start h-full">
```
wpf500: Sorry I didn't see this before but I think it might be better to apply `flex flex-col items-start` to all breakpoints (i.e. remove `lg:`). It probably doesn't matter in this case but flex affects things like collapsing margins so it's good practice to apply to non-conditionally where possible (or wrap the inner elements in a div so they aren't treated as flex children)

src/components/pages/callouts/CalloutIntroPanel.vue
```diff
@@ -0,0 +1,27 @@
...
+      <AppTitle class="pt-8 mb-4">{{ callout.title }}</AppTitle>
+      <figure class="mb-6">
+        <img class="w-full object-cover" :src="callout.image" />
+      </figure>
+      <div class="mb-6 text-lg" v-html="callout.intro"></div>
+      <AppButton variant="primary" class="px-6" @click="$emit('close')">
+        {{ t('callout.getStarted') }}
```
wpf500: Could you change this to `actions.getStarted` as generally I'm trying to put all generic button actions in the `actions` namespace. I've already added it to the locale spreadsheet for another card so you just need to update the key here

wpf500 (approved):

PR #496, opened 2023-10-19
feat: Login with 2FA using OTP

JumpLink:
## Changes

* **New slider component** - Currently, I've added a new slider component, which was well-suited to help me get back into Vue 3. This component will be needed in this PR for 3 steps to activate the 2FA but is not implemented yet.
* **Improved `CalloutStepper` step types** - I am considering using the `AppStepper` for the 2FA steps, I noticed missing types which I fixed.
* **New QR code component** - We also need a QR code generator for the 2FA process
* **Set MFA component** - New component to setup MFA TOTP
* Added button to contact page to disable MFA for admins
* Added some component to _themes for demonstrations
* Started to move types to `./src/type`

See https://github.com/beabee-communityrm/beabee/pull/319 for the backend changes of this PR.

### Screenshots

[MFA-flow.webm](https://github.com/beabee-communityrm/beabee-frontend/assets/1073989/75537b26-9a32-4a29-bf50-afa25841a5d1)

[disable-mfa-admin.webm](https://github.com/beabee-communityrm/beabee-frontend/assets/1073989/216ef40f-8a20-44dc-b9d7-9a7a048d8652)

[AppSlider-example.webm](https://github.com/beabee-communityrm/beabee-frontend/assets/1073989/bc34b06b-1cc8-46ae-9410-e56b925456c3)

AppQRCode-example

## Checklist before merging

- [x] Translations for all new i18n strings (except `ru` and `nl`)
- [x] 2FA can be set up
- [x] User can login with 2FA enabled
- [x] The admin can disable 2FA
- [x] The user can't login with 2FA and a invalid token


src/pages/_theme.vue
```diff
@@ -5,59 +5,159 @@ meta:
...
+    </section>
+  </section>
 </template>
 
 <script lang="ts" setup>
 import AppColor from '../components/AppColor.vue';
+import AppButton from '../components/button/AppButton.vue';
+import AppSlider from '../components/AppSlider.vue';
+import {
+  faAngleRight as NextIcon,
```
wpf500: Not a big thing but I think it's a bit clearer and more consistent with the rest of the codebase to reference these directly rather than creating an alias

src/components/AppSlider.vue
```diff
@@ -0,0 +1,140 @@
...
+const emit = defineEmits<{
+  (e: 'slide', details: AppSliderSlideEventDetails): void;
+}>();
+
+const props = withDefaults(defineProps<AppSliderProps>(), {
+  infinite: false,
+});
+
+const activeSlide = ref(0);
+const slideCount = ref(0);
```
wpf500: This should be a `computed`

```ts
const slideCount = computed(() => slideEls.value.length)
```

src/components/AppSlider.vue
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Handle window resize events
+ */
+const handleResize = () => {
+  // If the user resizes the window, we want to make sure the active slide is still centered
+  toSlide(activeSlide.value, 'auto');
+};
+
+onMounted(() => {
+  slideEls.value.push(
```
wpf500: Why push over assignment? If you use push you might need to clear the array in `onUnmounted` to avoid weird state, I think there might be some scenarios where a component can be remounted without being reinitialised.

```ts
slideEls.value = Array.from(...)
```

src/components/AppSlider.vue
wpf500, line 1: Looks really good and nice to see documentation :tada: .

My only hesitation is around the DOM references, I would generally try to avoid these where possible. I wonder if a (slightly) cleaner implementation could be to use transforms and transitions to do the scrolling, but without needing a reference to the slide element. You'd still need a reference to the container to see how many slides there are, but the scrolling would just be done with `transform: translateX(-100% * slideNo)`. This also has the added benefit of not needing to listen for the window resize event.

If you stick with this implementation then I don't think you should emit the `slideEl` in the slide event, so as not to encourage using DOM references.

src/pages/_theme.vue
```diff
@@ -5,59 +5,159 @@ meta:
...
+    </section>
+
+    <section class="my-5">
+      <h3 class="mb-2 font-title text-xl">AppSlider</h3>
+      <div class="bg-white p-4">
+        <div class="max-w-[1200px] mx-auto relative">
+          <AppSlider :show-navigation-buttons="true" :infinite="true">
+            <!-- Put your slides here -->
+            <template #slides>
+              <div class="flex-none w-full bg-danger">
```
wpf500: I wonder if a `AppSliderSlide` (or a better name) would be helpful here, as I guess `flex-none w-full` will be needed on all slides. It might be a bit overkill though?

```vue
<AppSlider ...>
  <AppSliderSlide></AppSliderSlide>
  <AppSliderSlide></AppSliderSlide>
  <AppSliderSlide></AppSliderSlide>
</AppSlider>

File: src/pages/_theme.vue [hunk 5]
```diff
@@ -5,59 +5,159 @@ meta:
...
-      <AppColor name="bg-grey-lighter" shade="Lighter" />
-      <AppColor name="bg-grey-light" shade="Light" />
-      <AppColor name="bg-grey-dark" shade="Dark" />
-      <AppColor name="bg-grey-darker" shade="Darker" />
-    </div>
-  </div>
+  </section>
+
+  <section class="my-5">
+    <h2 class="mb-2 font-title text-2xl">Components</h2>
```
wpf500, line 63: Thanks for starting this :+1:

src/pages/_theme.vue
JumpLink: Good idea

src/components/AppSlider.vue
JumpLink: I like the implementation using a native browser scrollbar, because it's also very easy to implement draggscroll and it is fast in its execution. But you are right, most slider implementations in the wild use CSS `transform` for this, maybe for a good reason. One reason could be that the scroll animation can't be changed, because it's always the scroll animation speed of the platform and browser. On the other hand, the user has his usual scrolling behaviour of his platform (especially on mobile devices).

I have implemented the sliders of https://www.seeberger.de/ (for example the product cards on the home page) in a similar way and I am very happy with it. The only disadvantage I had so far was that for (older?) Safari versions I needed the following polyfill to get a scrolling animation working: https://github.com/iamdustan/smoothscroll

src/components/AppSlider.vue
JumpLink: under another framework I had the problem that this way the complete array was re-rendered every time (for example if there is an HTML element for each entry of the array like I would have with the `v-for` attribute in Vue) instead of just the value that changed, I don't know how this behaves under Vue though, it probably won't make a noticeable difference, so I'm happy to change this

src/components/AppSlider.vue
JumpLink: thanks

src/pages/_theme.vue
JumpLink: I did it this way because I could easily test other icons and only had to change the code in one place, but I can adapt it if you like.

wpf500: I like the idea but I think for now let's just do simple imports for now. It would be good to have another file that maps Font Awesome icons to our "vocabulary", I'll add a task to the backlog.

src/components/AppSlider.vue
wpf500: Yeah fair enough, let's stick with the browser scrollbar :+1:

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -0,0 +1,371 @@
...
+    {{ t('accountPage.mfa.title') }}
+  </AppHeading>
+
+  <AppButton
+    v-if="isEnabled"
+    variant="primaryOutlined"
+    :icon="faMobileAlt"
+    @click="disableMFA"
+  >
+    {{ t(`accountPage.mfa.disable`) }}
```
wpf500: The i18n sheet is a bit of a mess but I'm trying to namespace reusable button actions under `actions.*` so we don't end up with lots of duplicate `**.enable` entries. Can you move these and the other common actions to `actions.*`?  (you'll find them near the top of the sheet)

i.e. here it would be `actions.disable` rather than `accountPage.mfa.disable`

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -0,0 +1,371 @@
...
+  </AppButton>
+
+  <AppButton
+    v-else
+    variant="primaryOutlined"
+    :icon="faMobileAlt"
+    @click="toggleMFAModal"
+  >
+    {{ t(`accountPage.mfa.enable`) }}
```
wpf500: `actions.enable`

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -0,0 +1,371 @@
...
+          <div
+            class="whitespace-break-spaces h-full flex flex-col justify-between items-center"
+          >
+            <p class="text-center">
+              {{ t(`accountPage.mfa.enterCode.desc`) }}
+            </p>
+            <span class="w-full h-full flex flex-col justify-center px-4">
+              <AppInput
+                v-model="userToken"
+                :value="userToken"
```
wpf500: I don't think you should need this, should be handled by v-model?

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -0,0 +1,371 @@
...
+      <template
+        #navigation="{ nextSlide, prevSlide, isFirstSlide, isLastSlide }"
+      >
+        <span class="flex justify-between mt-3">
+          <AppButton
+            v-if="isFirstSlide"
+            variant="linkOutlined"
+            @click="closeMFAModal()"
+          >
+            {{ t(`accountPage.mfa.closeButton.label`) }}
```
wpf500: `actions.close`

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -0,0 +1,371 @@
...
+          </AppButton>
+          <AppButton v-else variant="linkOutlined" @click="prevSlide()">
+            {{ t(`accountPage.mfa.prevButton.label`) }}
```
wpf500: `actions.back`

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -0,0 +1,371 @@
...
+          </AppButton>
+          <AppButton
+            v-if="isLastSlide"
+            :disabled="!allStepsDone"
+            variant="link"
+            @click="saveMFA()"
+          >
+            {{ t(`accountPage.mfa.saveButton.label`) }}
```
wpf500: `actions.save`

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -0,0 +1,371 @@
...
+          </AppButton>
+          <AppButton v-else variant="link" @click="nextSlide()">
+            {{ t(`accountPage.mfa.nextButton.label`) }}
```
wpf500: `actions.next`

src/utils/api/api.interface.ts
```diff
@@ -52,6 +52,32 @@ interface ContactData {
...
+/**
+ * Contact multi factor authentication data
+ * TODO: Move to common
+ */
+interface ContactMfaData {
+  secret?: string;
+  type: ContactMfaType;
+}
+
+export type fetchContactMfaData = Pick<ContactMfaData, 'type'> | null;
```
wpf500: Should be `GetContactMfaData`

src/utils/api/contact-mfa.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Fetch a contact MFA
+ * @param contactId The contact id
+ * @returns
+ */
+export async function fetchContactMfa(
+  contactId: string
+): Promise<fetchContactMfaData> {
+  let { data } = await axios.get<fetchContactMfaData>(
```
wpf500: Should be `... axios.get<Serial<GetContactMfaData>>`.

`Serial` is designed to help mitigate against forgetting to deserialize data from the API, so if in the future you changed `GetContactMfaData` to:

```ts
interface GetContactMfaData {
  ...
  date: Date;
}
```

The compiler would throw an error saying `Serial<GetContactMfaData>` isn't the same as `GetContactMfaData` (because the date needs to be converted from a string to a Date)

src/utils/api/contact-mfa.ts
```diff
@@ -0,0 +1,56 @@
...
+  CreateContactMfaData,
+  Serial,
+} from './api.interface';
+
+/**
+ * Deserialize a contact MFA
+ * @param data The data to deserialize
+ * @returns The deserialized data
+ */
+export function deserializeContactMfa<T>(data: T): T {
```
wpf500: Should the definition not be as follows?
```ts
export function deserializeContactMfa(data: Serial<GetContactMfaData>): GetContactMfaData {
  ...
}
```

src/utils/api/contact-mfa.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Create a new contact MFA
+ * @param contactId The contact id
+ * @param dataIn The data to create the contact MFA
+ * @returns
+ */
+export async function createContactMfa(
+  contactId: string,
+  dataIn: CreateContactMfaData
+): Promise<CreateContactMfaData> {
```
wpf500: Does this not just return `GetContactMfaData`? (i.e. without the secret and token?)

src/utils/api/contact-mfa.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @param contactId The contact id
+ * @returns
+ */
+export async function fetchContactMfa(
+  contactId: string
+): Promise<fetchContactMfaData> {
+  let { data } = await axios.get<fetchContactMfaData>(
+    `/contact/${contactId}/mfa`
+  );
+  data = data || null;
```
wpf500: Not sure what this is for?

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -0,0 +1,371 @@
...
+  resetState();
+};
+
+/** Close the modal */
+const closeMFAModal = () => {
+  showMFASettingsModal.value = false;
+};
+
+/** Toggle the modal */
+const toggleMFAModal = () => {
```
wpf500: I tend to just inline small functions like this as it feels just as readable, do you think it's better to have them explicitly defined?

src/components/AppQRCode.vue
```diff
@@ -0,0 +1,98 @@
...
+  qr.make();
+
+  qrContainerEl.value.innerHTML = qr.createSvgTag(4);
+  const svgEl = qrContainerEl.value.children[0] as SVGElement;
+  svgEl.setAttribute('width', '100%');
+  svgEl.setAttribute('height', '100%');
+};
+
+onMounted(() => {
+  watch(props, onQrDataChanged, { deep: true });
```
wpf500: Is there a reason this is in `onMounted` rather than just in the main scope? I think there are cases where this could result in double watching as the component can be remounted without being reinitialised.

Better to put it outside `onMounted` I think

src/pages/auth/login.vue
```diff
@@ -80,21 +105,41 @@ const loading = ref(false);
...
+
+let emailBefore = '';
 
 async function submitLogin() {
   loading.value = true;
   hasCredentialError.value = false;
+  hasWrongMFAToken.value = false;
+
+  // Only reset MFA if email changed
+  if (emailBefore !== data.email) {
```
wpf500: I think this could be a `watch` so it resets when the user changes the input and you don't need the `emailBefore` variable, maybe something like
 
 ```ts
 watch(() => data.email, () => (hasMFAEnabled.value = false));
 ```

src/pages/auth/login.vue
```diff
@@ -80,21 +105,41 @@ const loading = ref(false);
...
+  emailBefore = data.email;
   try {
     await login(data);
     await updateCurrentUser();
     // TODO: use router when legacy app is gone
     window.location.href = isInternalUrl(redirectTo) ? redirectTo : '/';
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
   } catch (err: any) {
-    if (err.response?.status === 401) hasCredentialError.value = true;
+    if (err.response?.status === 401) {
```
wpf500: You could use `isApiError(err, code)` here, it handles checking the type so `err` can be `unknown` instead.

It currently only checks for 400 status code, but you could change the signature to
```ts
function isApiError(err: unknown, code: string, status?: number = 400)
```
and update accordingly.

src/pages/_theme.vue
```diff
@@ -5,59 +5,246 @@ meta:
...
+
+    <section class="my-5">
+      <h3 class="mb-2 font-title text-xl">AppQRCode</h3>
+      <div class="bg-white p-4 w-64">
+        <AppQRCode :qr-data="appQRCode.data" />
+
+        <p>Change the data to see the QR code change.</p>
+        <AppInput
+          v-model="appQRCode.data"
+          :value="appQRCode.data"
```
wpf500: I think you don't need this?

src/utils/api/api.interface.ts
```diff
@@ -52,6 +52,32 @@ interface ContactData {
...
+  TOTP = 'totp',
+  // E.g. U2F, EMAIL, SMS, HOTP, etc.
+}
+
+/**
+ * Contact multi factor authentication data
+ * TODO: Move to common
+ */
+interface ContactMfaData {
+  secret?: string;
```
wpf500: Why define `secret` here at all?

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -0,0 +1,371 @@
...
+  appStepper.value.steps.forEach((step) => {
+    step.validated = false;
+    step.error = false;
+  });
+  userToken.value = '';
+  userTokenValid.value = false;
+};
+
+/** Sync the stepper with the slider */
+const syncStepperWithSlider = (details: AppSliderSlideEventDetails) => {
```
wpf500: This feels like useful common functionality. Could you move the stepper inside the slider component, maybe with an optional param for steps so you can use the slider without the stepper? I think if it's internal they will also be able to use the same state for active step index so you don't need any syncing code

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -0,0 +1,371 @@
...
+  totp = new TOTP({
+    issuer: newValue.issuer,
+    label: newValue.label,
+    secret: totpSecret.value,
+  });
+  totpUrl.value = totp.toString();
+};
+
+/** Validate the totp token / user input code */
+const validateTOTOToken = (window = 2) => {
```
wpf500: Typo, should be `validateTOTPToken`

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -0,0 +1,371 @@
...
+  resultStep.error = !userTokenValid.value;
+  validateStep.validated = userTokenValid.value;
+  resultStep.validated = userTokenValid.value;
+
+  return userTokenValid.value;
+};
+
+/** Are all steps done with no errors? */
+const allStepsDone = computed(() => {
+  return (
```
wpf500, line 545: Simpler with `appStepper.value.steps.every(s => s.validated && !s.error)`?

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -0,0 +1,371 @@
...
+    !appStepper.value.steps[2].error
+  );
+});
+
+/** Fetch the contact and set the totp identity */
+watch(
+  toRef(props, 'contactId'),
+  async (contactId) => {
+    const contact = await fetchContact(contactId, ['profile']);
+    totpIdentity.value.issuer = 'beabee'; // TODO: Use name of beabee instance
```
wpf500: You should be able to use `generalContent.value.organisationName` for this (or something like that)

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -0,0 +1,371 @@
...
+
+/** Validate the current step */
+const validateStep = (details: AppSliderSlideEventDetails) => {
+  if (details.slideNumber === 2) {
+    validateTOTOToken();
+  }
+};
+
+/** Validate all previous steps */
+const validatePreviousSteps = () => {
```
wpf500: There *should* be a much nicer way to do validation here, but I know this is also how the `CalloutStepper` does it.

In most places we use the vuelidate library and it has some really nice features that would almost certainly eliminate the need for this sort of thing, but I'm not sure how right now. Let's leave it like this for now then when I refactor `CalloutStepper` I might have a little look at this too.

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -0,0 +1,371 @@
...
+    />
+    <AppSlider ref="appSliderCo" @slide="onSlideChange">
+      <template #slides>
+        <AppSlide>
+          <div class="whitespace-break-spaces">
+            <p class="text-center">
+              {{ t(`accountPage.mfa.scan.desc`) }}
+            </p>
+            <AppQRCode v-if="totpUrl" :qr-data="totpUrl" />
+            <!-- TODO: Allow other options like show url / secret key here? -->
```
wpf500: As a rule you should always give the option to copy the secret key directly I think, for those who want to add the code to managed service on their computer like LastPass. Can you just add it as a quick readonly input?

wpf500 (changes requested):
Again looks really good :tada:!. A few comments but feel free to challenge any of them if you don't think it's the right way to go.

src/components/pages/profile/account/SetMFA.vue
JumpLink: Good idea

src/components/pages/profile/account/SetMFA.vue
JumpLink: thanks

src/components/pages/profile/account/SetMFA.vue
JumpLink: yes

src/components/pages/profile/account/SetMFA.vue
JumpLink: I will take a closer look, in my case I actually only do a real validation in the last slide, in the previous ones it is more of a pseudo validation.

src/components/pages/profile/account/SetMFA.vue
JumpLink: Since the method is only used once, I have now done it inline, now

src/components/pages/profile/account/SetMFA.vue
JumpLink: thanks

src/utils/api/api.interface.ts
JumpLink: Is removed, now

src/utils/api/contact-mfa.ts
JumpLink: I think we do not need this

src/components/pages/profile/account/SetMFA.vue
JumpLink: Yes, works

src/components/AppQRCode.vue
JumpLink: No I change this

src/pages/auth/login.vue
JumpLink: I can't find `isApiError`, do you mean `isRequestError`? But this method just checks for `err.response?.status === 400` should this be `err.response?.status >= 400`?

src/utils/api/contact-mfa.ts
```diff
@@ -24,12 +26,11 @@ export function deserializeContactMfa<T>(data: T): T {
 export async function createContactMfa(
   contactId: string,
   dataIn: CreateContactMfaData
-): Promise<CreateContactMfaData> {
-  const { data } = await axios.post<Serial<CreateContactMfaData>>(
+): Promise<void> {
+  await axios.post<Serial<CreateContactMfaData>>(
```
wpf500: You can remove the `Serial<CreateContactMfaData>>` here as that's the returned data type which in this case is empty

src/pages/auth/login.vue
wpf500: Ah sorry I meant `isRequestError`. I think it would be better to add an optional third parameter to change the status code it checks for, e.g.
```ts
function isRequestError(err: unknown, code: string, status?: number  = 400) {
...
}
```

wpf500 (changes requested):
Woo looking great, just a couple of quick change request.s

src/utils/api/contact-mfa.ts
JumpLink: Check

wpf500 (approved):

PR #501, opened 2023-10-26
feat: initial UI toolkit

aiscarvalho:
## Feat/initial UI toolkit

Created the Typography page to kickstart beabee's UI Toolkit (see page in `/admin/ui-toolkit`).
In this page we documented the existing heading components and body text styles to organize information hierarchically.

image

## Changes

While documenting the existing text styles we:
- adjusted the sizes for headings (larger text sizes), font weights, letter spacing and color
- updated text sizes and line height settings in tailwing config
- added a new component for h4 headings
- applied title component to callout card
- adjusted heading margins in user contribution page
- larger text size for the Tabs component in large screens (above 1280px)
- applied AppInfoList component to user dashboard page
- minor: fixed the title alignment in the Join Page

User dashboard Before/After 
image image


src/components/pages/profile/ContributionInfo.vue
```diff
@@ -1,26 +1,19 @@
...
-      <div class="title uppercase">{{ t('common.contributing') }}</div>
-
-      <div class="content">
-        {{ n(contact.contributionAmount, 'currency') }}<br />
-        {{ t('common.every') }}<br />
-        {{ period }}<br />
-      </div>
-    </div>
+<template
+  ><div class="flex pt-3">
```
wpf500: Needs a new line after the `<template>` here

src/components/PageTitle.vue
```diff
@@ -1,8 +1,11 @@
...
-  <div class="mb-5" :class="{ 'border-b border-primary-40 pb-3': border }">
-    <div class="flex justify-between items-center">
+  <div class="mb-5">
+    <div
+      class="flex justify-between items-center"
+      :class="{ 'border-b border-primary-40 pb-3': border }"
+    >
       <AppTitle
-        class="mr-4 text-primary-80"
+        class="mr-4 text-primary-80 !font-light"
```
wpf500: It would be better to use a prop here on `AppTitle` rather than adding style override

src/components/AppTitle.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <h1
-    class="font-title font-bold"
-    :class="big ? 'text-3.5xl md:text-5xl' : 'text-2.5xl'"
+    class="font-title font-semibold tracking-tight"
+    :class="big ? 'md:text-5xl' : 'text-3.5xl'"
```
wpf500: I think the big version needs a default text size as currently it just has the base text size on mobile

image

src/components/AppSectionHeading.vue
```diff
@@ -0,0 +1,8 @@
+<template>
+  <h4 class="font-title text-xl font-regular tracking-tight text-body">
```
wpf500: `tracking-tight` seems very dependent on the font so probably wouldn't work across all the available fonts?

src/components/AppSubHeading.vue
```diff
@@ -1,5 +1,7 @@
 <template>
-  <h3 class="font-title text-lg font-semibold"><slot /></h3>
+  <h3 class="font-title text-1.8xl font-semibold tracking-tight text-body-60">
```
wpf500: See above for `tracking-tight`

src/components/PageTitle.vue
```diff
@@ -1,8 +1,11 @@
 <template>
-  <div class="mb-5" :class="{ 'border-b border-primary-40 pb-3': border }">
-    <div class="flex justify-between items-center">
+  <div class="mb-5">
+    <div
+      class="flex justify-between items-center"
+      :class="{ 'border-b border-primary-40 pb-3': border }"
```
wpf500: I think moving the border changes the appearance for a `PageTitle` with a border and a subtitle, is this intentional?

Before:
image

After:
image

In practice no where actually uses subtitle and border together and only the account page uses the subtitle, so maybe we could just remove the subtitle functionality and just add it as a normal `<p>` on the account page?

src/components/callout/CalloutSummary.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div>
     <div class="mb-4 flex">
-      <div class="flex-1">
+      <div class="flex-1 space-y-3">
```
wpf500: Better to just add mb-3 to the AppSubHeading here as `space-*` is a bit of a [hacky utility](https://tailwindcss.com/docs/space#limitations)

tailwind.config.js
```diff
@@ -37,7 +37,7 @@ module.exports = {
         DEFAULT: '3px',
       },
       lineHeight: {
-        tight: '1.3',
```
wpf500, line 40: This is a big change that changes the standard line height for the whole UI, was it intentional? Quite a few things look more squashed now

image
image

wpf500 (commented):
Some good improvements here, a few comments added.

I think it would be good if rather than having `mb-*` on the heading components if they instead defined the standard spacing within the component, so we don't end up with lots of spacing variation in the UI. What do you think?

src/components/PageTitle.vue
aiscarvalho: Yes, makes much more sense. Added a prop `light` to AppTitle that is now used in PageTitle.

src/components/AppTitle.vue
aiscarvalho: Thanks for catching this. Added a base text size now.

src/components/AppSectionHeading.vue
aiscarvalho: We tested this with all the available fonts and think it works well. 
But maybe we can discuss and review this one together, in Essen.

src/components/PageTitle.vue
aiscarvalho: We removed the subtitle functionality and add a ´<p>' in the account page.

Thanks for suggesting this sensible update. We wanted to add a border to this title and when we did we noticed it was the only page with extra text below the title (and above the border).

src/components/callout/CalloutSummary.vue
aiscarvalho: Done! Removed all `space-y-*`

tailwind.config.js
aiscarvalho: Agreed!

When reviewing the text sizes we noted that we had a custom value for `leading-tight`, larger than the default line-height. We changed the value but didn't know this class was applied to the `body` and affected all text elements.
We removed it from `body` and will use it only when it's appropriate.

aiscarvalho:
@wpf500 thank you for the careful revision. We address most of the questions and suggest we discuss the titles tracking together, next week.

wpf500:
This PR has some great changes in it that I want to merge before we get lots of merge conflicts, so I've temporarily removed `tracking-tight` until we discuss it.

src/components/PageTitle.vue
```diff
@@ -1,26 +1,31 @@
...
 
 defineProps<{
   title: string;
-  subTitle?: string;
   border?: boolean;
   noCollapse?: boolean;
 }>();
 </script>
+<style lang="postcss" scoped>
+.has-border h1 {
```
wpf500: Whenever you want to create classes (especially ones with non-class-only selectors) there is probably a more "Tailwindy" way. In this case it's actually much simpler just to remove `pb-3` from the border then you don't need to do anything to AppTitle

Issue #90, opened 2022-01-21
Numbers are cut off at bottom 


saschafoerster:
Somehow the numbers are not fully visible. 
<img width="449" alt="Bildschirmfoto 2022-01-21 um 17 25 36" src="https://user-images.githubusercontent.com/1153741/150563352-f2015f61-692f-494a-bdc5-ec956493a067.png">

saschafoerster:
Same in the backend  ./profile/contribution

<img width="396" alt="Bildschirmfoto 2022-01-21 um 17 27 47" src="https://user-images.githubusercontent.com/1153741/150563693-10d77d1e-6255-44ea-ac54-84cecb5acf10.png">

basilesimon:
Seems to me that the `<div.relative>` containing the number and its shadow `input` aren't dealt with very well in Safari (the only browser I can reproduce this bug on).

Very slightly tweaking the line-height of the `<div.text-6xl>` helps but does not feel very tailwind-y.
Alternatively, adding a bit of padding-bottom on the `<div.relative>` causes the baseline to go out of the window, and the number to hang a bit... which we may or may not be fine with.

saschafoerster:
Indeed, I saw this in Safari, sorry for not describing it in more detail before! 

PS: Isn't it that Safari users pay more? 😇

wpf500:
Fixed by #504!

PR #504, opened 2023-10-27
feat: add optional contribution option

wpf500:
Added the optional contribution check box to the join page and the setting to enable it to the membership builder.

I also fixed the long standing #90 which was only reported in Jan 2021 :wink:, and abstracted the contribution period/amount selector into a new generic `AppChoice`.

### Screenshots

image

## Checklist before merging


src/components/forms/AppInput.vue
```diff
@@ -80,6 +83,7 @@ const props = withDefaults(
     sameAs: undefined,
     pattern: undefined,
     prefix: undefined,
+    suffix: undefined,
```
JumpLink, line 86: Why are the values here initialised with `undefined`?

src/components/contribution/ContributionAmount.vue
```diff
@@ -116,7 +118,9 @@ const hasError = computed(() => validation.value.$errors.length > 0);
 
 const rules = computed(() => ({
   amount: {
-    minValue: minValue(toRefs(props).minAmount),
+    ...(!props.disabled && {
+      minValue: minValue(toRefs(props).minAmount),
+    }),
   },
 }));
```
JumpLink: I find this difficult to read, why not instead like this?

```ts
cconst ruls = computed(() => {
  if (props.disabled) return { amount: {} };

  return {
    amount: {
      minValue: minValue(toRefs(props).minAmount),
    },
  };
});
```

JumpLink (approved):
is approved, just two beauty questions / suggestions

src/components/forms/AppInput.vue
wpf500: It seems strange to me too but if I don't set a default on optional props then I get an ESLint warning:

image

src/components/contribution/ContributionAmount.vue
wpf500: I'm definitely guilty of doing this sort of thing way too much :+1:

What do you think of using a ternary operator, e.g.
```ts
const rules = computed(() =>
  props.disabled
    ? { amount: {} }
    : {
        amount: {
          minValue: minValue(toRefs(props).minAmount),
        },
      }
);
```

I find ternary operators easy to read and offer less potential for future developers to do strange things with not seeing early returns etc.

JumpLink: Looks good 👍

src/components/forms/AppInput.vue
JumpLink: okay

PR #510, opened 2023-11-03
feat!: API keys with expiry dates

wpf500:
Allow admins to create API keys with an expiry date to make them (slightly) more secure.

### Changes

* **Improved `AppCheckbox`** - made `modelValue` optional, this is useful for cases where you want to require the user to confirm an action, e.g.
  ```vue
  <AppCheckbox label="Confirm you understand the risks" required />
  ``` 
* **New time component** - `AppTime` show the relative time to now with correct title and datetime props
  * Replaced lots of custom `formatDistanceLocale` code with this component
* **Fixed `Serial` type** - correctly convert `Date | null` to `string | null` instead of `string`
* **Improved typing on `AppTable` and `AppPaginatedTable`** - Use newly supported generics. This started because I wanted to use `rowClass` on AppPaginatedTable without `any`, and it seems that Volar now supports it!

See https://github.com/beabee-communityrm/beabee/pull/321 for backend changes.

### Screenshots

image

## Checklist before merging


src/components/table/table.interface.ts
```diff
@@ -1,3 +1,10 @@
+export interface Item {
+  selected?: boolean;
+  // TODO: Try applying proper types to this
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any;
```
JumpLink, line 5: In `AppPaginatedTable.vue` you have `generic="I extends Item"` so item should be extended and that `[key: string]: any;` is not necessary?

JumpLink, line 5: This is not necessary, as you can simply inherit from `Item` and then pass this type to the component?

src/components/table/AppTable.vue
```diff
@@ -82,7 +82,7 @@
   </table>
 </template>
 
-<script lang="ts" setup>
+<script lang="ts" setup generic="I extends Item">
```
JumpLink, line 85: Interesting feature, how would I pass a generic parameter to this component?

src/pages/admin/settings/api-keys.vue
```diff
@@ -80,6 +83,26 @@ meta:
...
+        <AppSelect
+          v-model="newApiKeyData.expiresInDays"
+          :label="t('apiKey.data.expires')"
+          :items="[
+            { id: 0, label: t('adminSettings.apikey.expires.never') },
+            ...[30, 60, 90].map((days) => ({
+              id: days,
+              label: t('adminSettings.apikey.expires.days', { n: n(days) }),
+            })),
+          ]"
```
JumpLink: Maybe like this instead?

```ts
const generateExpiresInOptions = () => {
  const neverExpiresOption = { id: 0, label: t('adminSettings.apikey.expires.never') };
  const daysOptions = [30, 60, 90].map(days => ({
    id: days,
    label: t('adminSettings.apikey.expires.days', { n: n(days) }),
  }));

  return [neverExpiresOption, ...daysOptions];
}

const expiresInOptions = computed(generateExpiresInOptions);
```


```html
  <AppSelect
    v-model="newApiKeyData.expiresInDays"
    :label="t('apiKey.data.expires')"
    :items="expiresInOptions"
    :info-message="t('adminSettings.apikey.expiresHelp')"
    required
    class="mb-4"
  />
  ```

src/components/table/table.interface.ts
wpf500: It's required to generically access properties on the item in `AppTable`:
https://github.com/beabee-communityrm/beabee-frontend/blob/cfd0564de8ca24b7a826e73e1dc103d7787cb4e6/src/components/table/AppTable.vue#L67

It would be really amazing to make this all type safe but it's way above my type skills so would take me too long at the moment!

JumpLink (approved):
Looks good to me, again only cosmetic comments

src/components/table/AppTable.vue
wpf500: As in when you use the component? I don't know how you explicitly pass types but for the most part it will infer them from your props

src/components/table/table.interface.ts
JumpLink: Okay

src/pages/admin/settings/api-keys.vue
wpf500: I knew you'd call me up on this :smile:... I've been coding on my own for too long! I'll fix it

src/components/table/AppTable.vue
JumpLink: good to know, I'll try it out when I get the chance, thanks

PR #511, opened 2023-11-03
feat!: Reset MFA via email

JumpLink:
## Overview
This pull request adds the necessary frontend components for the "Reset MFA via Email" feature. It provides users with a seamless flow to reset their multi-factor authentication (MFA) device through a secure email link if they have lost access to their device.

### What has been done
- **Lost Device Link**: Implemented a "Lost my device" link on the login page to initiate the MFA device reset process.
- **New Lost-Device Page/Route**: Created a new page and route (`/lost-device`) where users can request the MFA reset link. This page supports an optional email parameter to prefill the user's email address.
- **New Reset-Device Page/Route**: Added a new page and route (`/reset-device`) that users are directed to from the reset link in their email. Here, they can enter their password and complete the reset process.
- **Types**: Moved API types and interfaces to `./src/type`
- **ResetSecurityFlowService**: New service oriented on the server side `ResetSecurityFlowService`.

## Other

Backend PR: https://github.com/beabee-communityrm/beabee/pull/322

## Screenshots

grafik
grafik
grafik
grafik
grafik
grafik
grafik

## Notes for the review

@wpf500 I started to move some types to `./src/type`, for this I had to adapt many imports to this new path, in this step I updated all imports directly with the new import alias path for this file, unfortunately this led to the fact that the changes became very large. Most of the import adjustments are in a single commit, it might be easier to look at this commit separately.

## Checklist what should work
- [x] The user can reset the password via email
- [x] The user can reset the 2FA via email and need's to enter his password for that
- [x] The user can reset the password via email and need's to enter his 2FA one time password for that
- [x] The user should not be able to reset both at once

## Checklist before merging

## Should be tested

- [x] The user can reset the password via email
- [x] The user can only reset the password with a valid MFA token if MFA is enabled
- [x] The user can only reset MFA with a valid password
- [x] The user can not reset the password if he has already requestet an unfinished MFA reset
- [x] The user can not reset MFA if he has already requestet an unfinished password reset


src/components/button/AppButton.vue
```diff
@@ -95,6 +96,15 @@ const props = withDefaults(
   }
 );
 
+const innerButton: Ref<HTMLAnchorElement | HTMLButtonElement | null> =
```
wpf500: You can also give the types to the `ref` method (then you don't need to import `Ref`)

```ts
const innerButton = ref<HTMLAnchorElement | HTMLButtonElement | null>(null);
```

src/utils/api/reset-security-flow.service.ts
```diff
@@ -0,0 +1,85 @@
...
+  UpdateResetDeviceData,
+} from '@type';
+
+/**
+ * Service for handling reset password and reset device flows.
+ *
+ * This is the service that is used by the reset password and reset device pages.
+ * This service is oriented on the server side `ResetSecurityFlowService`.
+ */
+class ResetSecurityFlowService {
```
wpf500: Although this might be a better structure I think it would be better for consistency to use the same style as other API files for now (i.e. direct exports of methods).

I have thoughts about restructuring the API methods too so would be good to discuss before we change it.

src/pages/auth/login.vue
```diff
@@ -35,11 +35,25 @@ meta:
...
           :label="t('form.password')"
         />
       </div>
 
+      <div class="mb-4">
+        <router-link
+          class="text-sm underline"
+          :to="
+            data.email
+              ? '/auth/forgot-password/' + data.email
```
wpf500: I think the email should be a query parameter rather embedded in the route, e.g. `/auth/forgot-password?email=asdadas`. This way you don't need the extra route files (all the new index.vue files) and I also think it's just a bit more "HTTP like" (ultimately `/auth/forgot-password` and `/auth/forgot-password/<email>` are the same resource so should have the same path)

wpf500: I think you could also then just write this as:
```
:to="{path: '/auth/forgot-password', params: {email: data.email}}"
```
(not sure about exact syntax)

src/pages/auth/login.vue
```diff
@@ -53,9 +67,14 @@ meta:
...
       </div>
 
-      <div class="mb-4">
-        <router-link class="text-sm underline" to="/auth/forgot-password">
-          {{ t('login.forgotPassword') }}
+      <div v-if="hasMFAEnabled" class="mb-4">
+        <router-link
+          class="text-sm underline"
+          :to="
+            data.email ? '/auth/lost-device/' + data.email : '/auth/lost-device'
```
wpf500: Same as above

src/pages/auth/reset-password/[id].vue
```diff
@@ -139,7 +205,10 @@ async function handleSubmit() {
       router.push({ path: '/' });
     }
   } catch (err) {
+    onError(err);
     hasError.value = true;
+    loading.value = false;
```
Comment (wpf500) [inline src/pages/auth/reset-password/[id].vue]: Is this necessary as `loading.value` is reset on line 214 anyway (just don't do an early `return`)

src/pages/auth/reset-password/[id].vue
```diff
@@ -62,6 +77,27 @@ meta:
...
+        v-if="errorCode === RESET_SECURITY_FLOW_ERROR_CODE.MFA_TOKEN_REQUIRED"
+        variant="info"
+        class="mb-4"
+        :title="t('resetPassword.errorTitles.' + errorCode)"
+      >
+        <p>{{ t('resetPassword.errors.' + errorCode) }}</p>
+      </AppNotification>
+
+      <AppNotification
+        v-if="
```
Comment (wpf500) [inline src/pages/auth/reset-password/[id].vue]: Can be more simply written as
```vue
<AppNotification v-else-if="errorCode" ... />
```

src/pages/auth/reset-device/[id].vue
```diff
@@ -0,0 +1,149 @@
...
+      addNotification({
+        variant: 'success',
+        title: t('resetDevice.success'),
+      });
+      router.push({ path: '/' });
+    }
+  } catch (err) {
+    onError(err);
+    hasError.value = true;
+    loading.value = false;
```
Comment (wpf500) [inline src/pages/auth/reset-device/[id].vue]: Let it fall through to line 147?

src/components/pages/profile/account/SetMFA.vue
```diff
@@ -403,7 +427,7 @@ const disableMfaAndNotify = async () => {
 /** Called when an error occurs while creating MFA */
 const onCreateError = (error: unknown) => {
   if (
-    isRequestError<ContactMfaCreateApiErrorData>(error, undefined, 401) &&
+    isRequestError<ContactMfaCreateApiErrorData>(error, undefined, [401]) &&
     error.response.data.message === LOGIN_CODES.INVALID_TOKEN
```
wpf500: Why is this checking `data.message` but everywhere else it checks `data.code`?

wpf500 (commented):
Lots of great changes and tech debt clean up! A few comments :)

src/components/pages/profile/account/SetMFA.vue
JumpLink: Yes, there are still legacy issues from Passport, but I also set `code` in the backend, so this could already work with `code`

src/pages/auth/login.vue
JumpLink: that's what I was looking for but couldn't find, yes that's much better. There is a second template without email which simply imports the one with email, that can then probably also be omitted?

src/pages/auth/reset-password/[id].vue
Comment (JumpLink) [reply src/pages/auth/reset-password/[id].vue]: true, good eye 👍

src/pages/auth/reset-password/[id].vue
Comment (JumpLink) [reply src/pages/auth/reset-password/[id].vue]: good to know, thank you 👍

src/pages/auth/reset-device/[id].vue
Comment (JumpLink) [reply src/pages/auth/reset-device/[id].vue]: no, not if the error has been handled

src/utils/api/reset-security-flow.service.ts
JumpLink: I am a fan of classes with class names that express the model / service which then uses simple method names like create, get, etc..

For example:

```ts
const user = new UserService();
user.create(...)
```

but that's probably just a matter of taste and I can understand your argument that it should be consistent with the rest of the frontend.

And the style above would also work with loose methods:

```ts
import * as user from './api/user.ts';
user.create(...)
```


Do you already have ideas on how you would like to restructure the API methods and what you hope to achieve?

JumpLink: Classes or objects would have the advantage that we could define an interface or an abstract class that specifies what API methods should look like

wpf500: I really like the way the Stripe SDK is structured, broadly with everything namespaced by the resource type. I guess this is what you are saying around having an abstract class that defines the standard methods (get, list, delete, etc.)
```ts
stripe.customers.get()
stripe.invoices.get()
// etc.
```

The main reason I've left it as is for now is because I have a dream that this can all be [auto-generated](https://swagger.io/tools/swagger-codegen/) based on an OpenAPI spec (which is generated by our API controllers), then we can just do an import and not have to write any code ourselves

```ts
import beabee from @beabee/api

beabee.callouts.get(id) // or something!
```

So I'd like to look at this at the same time as looking into implementing an OpenAPI spec in general, as that will also auto-generate API docs as well.

src/pages/auth/login.vue
```diff
@@ -122,40 +117,26 @@ const { t } = useI18n();
...
-        hasMFAEnabled.value = true;
-      } else if (err.response?.data?.code === LOGIN_CODES.INVALID_TOKEN) {
-        hasWrongMFAToken.value = true;
-      } else {
-        hasCredentialError.value = true;
-      }
+    if (isRequestError(err, [LOGIN_CODES.REQUIRES_2FA], [401])) {
+      hasMFAEnabled.value = true;
+    } else {
+      throw err;
```
JumpLink, line 134: Missing `LOGIN_CODES.INVALID_TOKEN` check here? Also `hasWrongMFAToken` and `hasCredentialError` are not handled anymore. I think I was checking all `401` errors and if its not `REQUIRES_2FA` and not `INVALID_TOKEN`, it's another credential error

src/pages/auth/reset-device/[id].vue
```diff
@@ -29,69 +41,28 @@ meta:
...
 import AuthBox from '@components/AuthBox.vue';
-import AppNotification from '@components/AppNotification.vue';
 
 import { resetDeviceComplete } from '@utils/api/reset-security-flow';
 import { updateCurrentUser } from '@store/index';
 import { isInternalUrl } from '@utils/index';
-import { isRequestError } from '@utils/api';
 
-import { addNotification } from '@store/notifications';
+import AppForm from '@components/forms/AppForm.vue';
```
Comment (JumpLink) [inline src/pages/auth/reset-device/[id].vue]: If we want to keep the style I started, then this import should be below the import of the other components

JumpLink (commented):
Overall, it looks cleaner if we have more error handling done by the `AppForm`. But my feeling is that there will still be problems and that something may have been lost when we tested it. But that's just a feeling because I'm not so good at code reviews. But if the tests go well, it should be fine

src/utils/api/reset-security-flow.service.ts
JumpLink: Yes, something like that. In this point of view, it makes sense to keep the old structure until we have a new uniform solution. Then we don't mix different approaches, which is then confusing.

We should also consider building a beabee client sdk as a separate NPM package, which others - who want to use the beabee api - can then also use

src/pages/auth/login.vue
wpf500: See the definition of `<AppForm>` above. It should catch these error and handle showing the right message, so you only need to handle them yourself if you want to do something special

PR #522, opened 2023-12-13
feat: upgrade dependencies

wpf500:
Upgrades all frontend dependencies to their latest version :tada:

Most of the code changes come from:
* Prettier's new formatting for chained ternany statements
* Using `@vue/tsconfig` as a base introduced the flag [verbatimModuleSyntax](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax). This flag means you need to explicitly declare if you are importing a type or an actual value, which helps simplify the removing of imports (or not) when compiling


src/components/table/AppPaginatedTable.vue
```diff
@@ -58,4 +55,11 @@ defineProps<{
   selectable?: boolean;
   rowClass?: (item: I) => string;
 }>();
+
+// Slots are passed to AppTable, typing is currently lost
```
JumpLink, line 59: Why can you not use types here?

JumpLink (approved):
Looks good to me. Two quick questions, the first is in the code the second is: 

I prefer to import all type imports below all other imports. But we haven't introduced that as a rule, do we want to do that or do you think it's better this way?

src/components/table/AppPaginatedTable.vue
wpf500: Basically because of the way I am passing the slots down to `AppTable`.

`AppTable` has clearly defined slots, with a type that it something like
```ts
Partial<Record<`header-${string}`|`value-${string}`|"empty"|"loading", (_: {}) => any>>
```
This comes from the Vue compiler which sees that in `AppTable.vue` we create slots for those template strings.

However `AppPaginatedTable` just passes any slots it is given to `AppTable`, it doesn't specify `header-${string}` etc. So the list of slot names is any string, i.e.
```ts
const slotNames = Object.keys(slots);
// Type of slotNames is string[]
// But technically should be (`header-${string}`|`value-${string}`|"empty"|"loading")[]
```

The longer term fix is to figure out how to make `AppPaginatedTable` properly typed, but that's a bit beyond this PR :)

wpf500:
I agree about ordering the imports but I'd prefer to fix all the imports in one go by adding the Prettier import sorter rather than doing it piecemeal.

src/components/table/AppPaginatedTable.vue
JumpLink: okay thank you ;)

JumpLink:
@wpf500 Okay, from my side this can be merged 👍

PR #526, opened 2024-01-11
chore(style): Rename Content type names for a better overview / auto completion

JumpLink:
See screenshot:

Bildschirmfoto vom 2024-01-11 18-49-12

I copied this types to the beabee-client in the telegram-bot repository and changed them and have adapted them as in this PR.


wpf500 (approved):

JumpLink:
Since we are now using the same types in two projects, does it make sense to move it to common?

wpf500:
I guess it might make sense in the interim. I'm still holding out for these types to be generated by the OpenAPI spec and live in a new package called `beabee-client`, but that could be a way off yet.

PR #534, opened 2024-02-08
feat: Mobile add button

aiscarvalho:
## New floating add button in mobile

The add button, in mobile, is displayed in the bottom right corner as a floating button.
Added a new button size (large).

### Screenshots

Before/After
image
image


wpf500:
This is fine for now but repeated patterns should really be abstracted into new components rather than being copy and pasted. Hopefully we can revisit this soon

PR #546, opened 2024-02-22
Callout map: Add New review

rlafuente:
**Note:** we're still missing the mobile UI for this, but it's good for testing on desktop

- update Add New button placement
- Cancel option when picking a point on the map
- Add New button included in the gallery, which switches back into the map view and starts Add New mode
- use query parameters to remove intro loading, needed for switching between map and gallery **(is this the best way?)**
- use the `callout.addNew` existing i18n key


aiscarvalho:
We've just updated the branch with the mobile Ui featuring the "Add new location" button in a bottom bar.
The bottom bar is visible if the map is embed. We figured this made sense because the "Add new" is placed there but we wanted to double check with you.

The "Add new" button is centred since, for now, it's the only element displayed in the bottom bar.

Screenshot of the mobile UI and spec for reference:

image


image

src/pages/callouts/[id]/map.vue
```diff
@@ -114,19 +128,41 @@ meta:
...
             <font-awesome-icon :icon="faInfoCircle" class="mr-1" />
-            {{ t('callout.addAPoint') }}
+            {{ t('callout.addAPoint') }}.
+            <span class="cursor-pointer underline" @click="handleCancelAddMode"
+              >{{ t('actions.cancel') }}?</span
+            >
           </p>
         </div>
       </transition>
       <button
```
Comment (wpf500) [inline src/pages/callouts/[id]/map.vue]: This could now be an `AppButton` like the other examples so the styles are shared

wpf500 (changes requested):
A couple of bugs too:

Double image on mobile
image

Style problems between gallery and map:
[Screencast from 20-03-24 13:17:31.webm](https://github.com/beabee-communityrm/beabee-frontend/assets/2084823/81f084f8-aeca-4b8f-9427-6457e854dd40)

wpf500:
Thanks for doing this so quickly, just a couple of little things to fix

rlafuente:
Okay just pushed the fixes; one outstanding detail is that the Add button disappears when viewing the gallery on mobile. I can look closer into this tomorrow.

wpf500 (approved):

PR #550, opened 2024-02-23
feat!: multi-lingual callouts

wpf500:
Adds the ability to specify some text (title, intro, etc.) in multiple languages. The user interface is very simple for now, it just duplicates the input field for each language.

Note this does not enable multi-lingual inputs in the callout form, that is coming next

image

image


src/lib/i18n.ts
```diff
@@ -53,6 +53,8 @@ const i18n = createI18n({
   },
 });
 
+const userOnlyLocales = ['pt', 'ru'];
```
JumpLink: I think it would be cleaner if you added a new property in the locales object for it

JumpLink (approved):
All in all only a small change, so I can approve it. But I would like to see what translations are available in the `locales` directory when bootstrapping the app, so that no manual changes to the code are necessary for new languages.

*Edit*
Sorry, I'm not fully awake yet and have only looked at the last changes and will have a look at the rest

JumpLink (approved):
Looks good to me, great how you added multilingualism without breaking the backwards compatibility

PR #551, opened 2024-02-29
feat: add Captcha support with CaptchaFox

wpf500:
Add Captcha support for when submitting callout responses

image


src/components/pages/admin/callouts/callouts.interface.ts
```diff
@@ -39,6 +39,7 @@ export interface TitleAndImageStepProps {
 export interface SettingsStepProps {
   whoCanTakePart: 'members' | 'everyone';
   allowAnonymousResponses: 'none' | 'guests' | 'all';
+  requireCaptcha: 'none' | 'guest' | 'all';
```
JumpLink, line 42: Does it make sense to move the enum from the backend to common and use it here? Then it only needs to be used in one place

src/utils/api/callout.ts
```diff
@@ -108,13 +108,18 @@ export async function fetchResponsesForMap(
...
-    guestEmail: data.guestEmail,
-  });
+  await instance.post(
+    `/callout/${slug}/responses`,
+    {
+      answers: data.answers,
+      guestName: data.guestName,
+      guestEmail: data.guestEmail,
+    },
+    { params: { captchaToken } }
```
JumpLink, line 121: It is no problem to send the parameter even if it is not required?

JumpLink (approved):
Looks good, just two questions

src/utils/api/callout.ts
wpf500: If `captchaToken` is undefined it won't actually send the parameter I think, although I haven't actually checked that

src/components/pages/admin/callouts/callouts.interface.ts
wpf500: It does, I was just being lazy :upside_down_face:. I will fix

wpf500: I will fix this when updating beabee-common to your new version as the current one doesn't have any enums defined in it and seems pointless to add it when it's about to be removed in any case

PR #574, opened 2024-04-11
feat: add filtering by callout response for contacts

wpf500:
This PR adds the frontend support for filtering contacts by their answers to callouts, the related backend changes are in https://github.com/beabee-communityrm/beabee/pull/392.

image

`AppSearch*` components have been refactored to support dynamic filters. In this particular use case the filters are dependent on the callout the admin wants to filter by, when they choose a callout it loads the answer filters for that callout.

To support this filter groups can now define a custom component which handles rendering filter rules instead of the default component `AppSearchRuleFilterGroup`. See `CalloutResponseFilterGroup` for how this is implemented for callout response filtering.

This PR also includes the following changes:
*Added `type="button"` to `AppToggle` to stop it trying to submit forms
* Simplified `AppSearch` by moving most of the logic into new `AppSearchForm` and `AppSearchSummary` components which handle the form and summary views respectively
* `FilterGroup.items` used to be a `string[]` of filter item IDs, now it is the actual list of filter items. This simplifies the component logic as there's no need to pass both `filterGroups` and `filterItems` to the components. The method `withItem` gives developers a type-safe way to define filters for a given model.

The PR introduces a cleaner layout for filter groups, separating them out into "tabs" instead of optgroups

Before:
Screenshot from 2024-04-17 14-48-29

After:
image


JumpLink (approved):

src/assets/styles/multiselect.css
```diff
@@ -0,0 +1,33 @@
+@import '../../../node_modules/vue-multiselect/dist/vue-multiselect.css';
```
JumpLink: Would be better to change this to `@import 'vue-multiselect/dist/vue-multiselect.css';`

As preparation for the monorepo, nothing should point directly to the `node_modules` directory and instead be resolved via the module resolver of the build system. In this case it's `vite` which should be able to automatically resolves this.

JumpLink (changes requested):
Looks good so far, only the import of the styles should be changed

src/assets/styles/multiselect.css
wpf500: Ah thanks, have fixed

wpf500:
Closing in favour of https://github.com/beabee-communityrm/monorepo/pull/5

PR #598, opened 2024-05-16
fix: separate payment content out

wpf500:


JumpLink (approved):
Thank you, I would have done the same

